PR #3, opened 2018-11-30
Update AZTEC.sol

shogochiai:


ArnSch:
Thanks for this. Merging.

emkman:
@ArnSch I don't think this is correct and shouldn't have been merged. See:
https://medium.com/@chriseth/writing-smart-contracts-in-iulia-2a5ba737c7f1
It was originally called JULIA but a rename to IULIA was stared about a year ago, and now it actually called YUL apparently: https://solidity.readthedocs.io/en/develop/yul.html
JULIA seems to be the least correct option as it causes confusion with another language called Julia that has been around for some time.

ArnSch:
Thanks @emkman, thought that the latest development was a reversal from Iulia back to Julia and was mistaken.

I am tempted to keep it as Julia for the time being, given [docs](https://solidity.readthedocs.io/en/v0.4.23/julia.html) for v0.4.23 still name it as Julia and not YUL, or IULIA.

emkman:
No problem! Agreed that keeping it synced with the convention of the target
solidity version makes sense.

PR #8, opened 2019-01-03
Added linters, editorconfig and truffle ^5.0

PaulRBerg:
Major changes:

- Use [ethlint](https://www.ethlint.com/) (formerly solium). I added a lot of cool scripts, such as `lint`, `lint:sol`, `lint:sol:fix` etc.
- Solium and other important Ethereum devs like [OpenZeppelin](https://github.com/OpenZeppelin/openzeppelin-solidity/issues/152) use 2 space indenting.
- Use `.editorconfig` to enable interoperability and enforce rules across IDEs.
- Use the latest version of truffle. I wrote a [tutorial](https://paulrberg.com/post/2018/12/30/upgradeable-smart-contracts/) about this, and v5 is really cool. To keep using solidity 0.4.24 for the moment, I amended `trufle-config.json` to export a `compilers` configuration.

Other minor proposals:

- Use the simple form of the verbs when defining events ("Create" instead of "Created"). Also, ConsenSys' [best practices](https://consensys.github.io/smart-contract-best-practices/recommendations/#differentiate-functions-and-events) recommend prepending the event's name with `Log`.
- Only use `truffle-config.js` instead of `truffle.js` to get rid of technical debt (`truffle.js` breaks on Windows)
- Use the locally installed versions of `mocha` et al in `package.json`, much easier to configure on continuous integration services
- Rename scripts in `package.json` from "fooBar" to "foo:bar", easier to read for longer concatenations 

Notes:

- There is currently a false positive error when linting with solium. I [reported](https://github.com/duaraghav8/Ethlint/issues/248) it on their GitHub repo.
- Truffle ^5.0 outputs checksummed addresses in the `build` folder, but web3's `getCode` function does not, so tests began failing after I made the upgrade. I fixed this by converting the addresses to lowercase strings on line 32 of `aztecToken/controller.js`.


PaulRBerg:
Uhmm .. migrations, at least on Rinkeby, stopped working. If I do `truffle migrate --reset --verbose-rpc --network rinkeby` and a customly set private key, I get:

```
> {
>   "jsonrpc": "2.0",
>   "id": 3,
>   "method": "eth_getBlockByNumber",
>   "params": [
>     "latest",
>     false
>   ]
> }
```

So, for mysterious reasons, the migrations don't even start. I went back to the master branch and it worked correctly there.

PaulRBerg:
So I eventually solved the deployment issues by replacing `truffle-wallet-provider` with `truffle-hdwallet-provider` and the private keys with a mnemonic loaded via [dotenv](https://github.com/motdotla/dotenv). I even added a `.env.example` file to make it easy for new devs to set up their environment.

I tried to reconcile truffle v5 with Zac's private keys model, but I stumbled upon weird errors like `TypeError: e[t].replace is not a function` and it generally feels that truffle + infura + private keys is a [cumbersome configuration](https://github.com/trufflesuite/truffle/issues?utf8=✓&q=truffle+infura+private+key).

Nevertheless, I could continue trying to make it work with the previous model, if you insist on having it!

PaulRBerg:
**Update**: now you have to option to load a private key, but you can also set a mnemonic. The former takes precedence.

fvictorio:
Sorry for the unsolicited comment, but since you're adding tooling you may want to also add something for measuring the coverage of the contracts tests (I've used [solidity-coverage](https://www.npmjs.com/package/solidity-coverage) and it's good, but I think there are some alternatives).

build/contracts/AZTECERC20Bridge.json
```diff
@@ -8637,1518 +9097,1119 @@
...
+              "src": "7562:0:1"
+            },
+            "scope": 380,
+            "src": "7419:1288:1",
+            "stateMutability": "nonpayable",
+            "superFunction": null,
+            "visibility": "internal"
+          },
+          {
+
```
zac-williamson: Can we remove changes to the `build` folder from this pull request? The new contracts don't have deployed versions on mainNet and the test-nets. Until we've improved our deployment pipeline we need the contracts in `build` to point to their deployed counterparts.

PR #10, opened 2019-01-08
Fix link in README

fvictorio:


ArnSch:
Thanks for the fix @fvictorio, merging.

PR #31, opened 2019-01-30
Small edits to the build process

nicholaspai:
1. When testing AZTEC smart contracts against already deployed contracts for example on Rinkeby, I needed to instruct demoTransactions to use the deployed addresses.

This is because when I `truffle compile --all` in `packages/protocol`, the new Contract ABI's do not have "network" properties since I was not the one to deploy them.

Therefore, I imported `getContractAddressesForNework` into `demo/resources/contracts/[erc20Token/aztecToken` and added a fallback if statement to retrieve contract addresses from "@aztec/contract-addresses"

2. added `.env` to a `.gitignore` in `packages/protocol`

3. Had to switch my network_id in the development "network" JSON to "*"

## Packages Updated

# #Checklist

- [ ] Write tests for added code.
- [ ] Prefix PR title with [WIP] if necessary.
- [ ] Update documentation when needed.
- [ ] Add new entries to the relevant CHANGELOG.json files.


PaulRBerg:
Hey, many thanks for this 🙏

Unfortunately, after running your code, I realised that the fallback functionality cannot be merged because of two reasons:

1. When `truffle compile` is not executed, neither the `demo:*` nor the `test` scripts work anymore because the `build/contracts` folder is missing - and this is required as a module in the demo.
2. Even if the fallback would fix for the testnet demos, it would break the tests, because we don't have ganache contract addresses in `@aztec/contract-addresses`. To fix this, we'd have to take a [ganache snapshot](https://ethereum.stackexchange.com/questions/40153/how-can-i-take-a-snapshot-of-ganache-or-truffle-develop-state-and-reload-later).

We were planning to rewrite the demo as a standalone starter project anyway, so it's not worth it to make any hot-fixes now. However, the following changes are kept and thank you for suggesting them:

- Added `.env` to `.gitignore` - we normally put this in the global gitignore, but better have it here just in case!
- Replaced the development network id in `truffle-config.js` from `1234` to `*`

Closing this in favour of #33.

nicholaspai:
Thanks for looking over this Paul! All changes understood. Regarding
`build/contracts` folder missing, I did encounter this as well. I'd
recommend either adding `cd packages/protocol && truffle compile --all` to
the `demo` README, as this error was brought to my attention (and easily
fixed) when I tried to run `npm run demo:*` straight after cloning and
installing prerequisites!

PaulRBerg:
Good catch! I must've messed up the READMEs, because on the top-level monorepo page there is a step that compiles and deploys the contracts:

```bash
$ truffle migrate --network rinkeby
```

Will update the demo README accordingly. Thank you!

PR #37, opened 2019-02-04
Solc >0.5.0 + Openzeppelin >2.1.0

nepalbitcoin:
## Problem
Breaking changes in Solc >0.5.0.

## Solution
Update Openzeppelin to newer version and change contract codes to reflect solc > v0.5.0. Test codes runs OK.

## Packages Updated
"openzeppelin-solidity": "^2.1.0"

## Checklist

- [ ] Update documentation when needed.
- [ ] Add new entries to the relevant CHANGELOG.json files.

### DIFF
--- a/packages/protocol/contracts/ACE/ACE.sol
+++ b/packages/protocol/contracts/ACE/ACE.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.4.24;
+pragma solidity >=0.5.0 <0.6.0;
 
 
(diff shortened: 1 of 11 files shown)


ArnSch:
Hi @nepalbitcoin and @0xC0DE4C0FFEE thanks for contributing. We're doing some pretty significant changes in the next couple of weeks, and will be upgrading to 0.5.0 once we're a bit more stable!

PR #46, opened 2019-02-12
Feat LibEIP712

PaulRBerg:
Implements a modular design for handling EIP712 domain hashes and signatures. Forked from https://github.com/0xProject/0x-monorepo/blob/development/contracts/exchange-libs/contracts/src/LibEIP712.sol

- [x] Integrate in JoinSplit
- [x] Integrate in ZKERC20


packages/dev-utils/src/constants.js
```diff
@@ -62,10 +69,7 @@ module.exports = {
         },
         primaryType: 'ACE_NOTE_SIGNATURE',
     },
-    ACE_DOMAIN_PARAMS: {
-        name: 'AZTEC_CRYPTOGRAPHY_ENGINE',
-        version: '0.1.0',
-    },
+    ACE_NOTE_SIGNATURE_HASH: keccak256('NoteSignature(bytes32 noteHash,address spender,bool status)'),
```
zac-williamson: I think we can remove this - I can't see it being used anywhere and this type of hash is created by `eip712.encodeStruct`  when supplied with an EIP712  schema

PaulRBerg: Fixed in the last commit

PR #54, opened 2019-02-20
Feat SafeMath

PaulRBerg:
## Problem

Contracts do not use [SafeMath][safemath].

## Solution

Implement SafeMath in all contracts. Better merge #53 first, I branched off from there.

## Packages Updated

- protocol
- contract-artifacts

## Checklist

[safemath]: https://docs.openzeppelin.org/docs/math_safemath

### DIFF
--- a/packages/protocol/contracts/ZKERC20/ZKERC20.sol
+++ b/packages/protocol/contracts/ZKERC20/ZKERC20.sol
@@ -1,10 +1,13 @@
 pragma solidity >=0.5.0 <0.6.0;
 
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
 import "../ACE/ACE.sol";
 import "../libs/LibEIP712.sol";
 
 contract ZKERC20 is LibEIP712 {
     using NoteUtils for bytes;
+    using SafeMath for uint256;
     
     // Hash for the EIP712 Note Signature schema
     /* solhint-disable */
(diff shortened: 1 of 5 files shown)


packages/protocol/contracts/utils/IntegerUtils.sol
```diff
@@ -1,10 +1,13 @@
 pragma solidity >=0.5.0 <0.6.0;
 
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
 library IntegerUtils {
-    
+    using SafeMath for uint256;
+
     function uintToBytes(uint256 num, uint256 size) internal pure returns (bytes5 output) {
```
zac-williamson: TODO: we can make an efficient casting method for this, but we can put that in another PR

packages/protocol/contracts/ZKERC20/ZKERC20.sol
```diff
@@ -200,14 +202,14 @@ contract ZKERC20 {
     }
 
     function logInputNotes(bytes memory inputNotes) internal {
-        for (uint i = 0; i < inputNotes.getLength(); i += 1) {
+        for (uint i = 0; i < inputNotes.getLength(); i = i.add(1)) {
```
zac-williamson: TODO: we should discuss whether using safeMath for iterations like these is redundant

PR #55, opened 2019-02-25
Minor fixes to AZTEC demo README

cryptogoth:
## Problem

The documented join-split has an incorrect index into proofData that will prevent it from both constructing and validating.
The given `0x000` address for confidential DAI also needs to be changed to a valid Ethereum address.

## Solution

Fix the above index, provide a random example address so that
`constructJoinSplit` executes.

## Checklist


PaulRBerg:
Hey Paul, many thanks for the heads up. I updated the example to use the latest aztec.js API!

PR #64, opened 2019-03-05
Merge ACE and NoteRegistry into one contract

PaulRBerg:
## Problem
Having a separate contract for the note registry causes a lot of friction for developers to pick up AZTEC.

## Solution

Make NoteRegistry a set of internal functions within ACE.

## Packages Updated

- protocol

## Checklist


packages/protocol/contracts/ACE/ACE.sol
```diff
@@ -93,8 +124,8 @@ contract ACE {
             calldatacopy(add(memPtr, 0x104), _proofData, add(calldataload(_proofData), 0x20))
 
             // call our validator smart contract, and validate the call succeeded
-            
-            if iszero(staticcall(gas, validatorAddress, memPtr, add(calldataload(_proofData), 0x124), 0x00, 0x00)) {
+            switch iszero(staticcall(gas, validatorAddress, memPtr, add(calldataload(_proofData), 0x124), 0x00, 0x00))
```
zac-williamson: we can remove `iszero` and test for `case 0` in the switch statement (we know that the return value of `staticcall` is already a boolean)

packages/protocol/contracts/ACE/ACE.sol
```diff
@@ -184,38 +215,6 @@ contract ACE {
         }
     }
 
-    function createNoteRegistry(
```
zac-williamson: It would improve readability to make the logic surrounding note registries a discrete contract, that `ACE.sol` inherits from. I don't think any of the logic needs to change, creating a discrete contract means we can approach testing ACE in a more systematic manner.  
  
Regarding the structure of an inherited `NoteRegistry.sol`, this base contract needs to validate the legitimacy of proofs via `validateProofByHash`. This should be a virtual function inside `NoteRegistry.sol` that `ACE.sol` overrides.

PaulRBerg:
Closing this in favour of https://github.com/AztecProtocol/AZTEC/pull/68

PR #89, opened 2019-04-15
feat: public range proof

thomas-waite:
## Summary
In many use cases, it is important to be able to prove that a confidential digital asset has a value  greater than or less than a public integer. This has to be performed in zero-knowledge - that is the value of the original confidential digital asset is never revealed. We call this proof a public range proof.

This PR implements such a proof - providing the Javascript proof construction code and the smart contract proof validator, along with relevant ABI encoding functionality and tests.

## What I would like feedback on/where to focus
1) `(aztec.js/src/abiEncoder/inputCoder.js)`: To make the API for `publicRange.encodePublicRangeTransaction()` general purpose for both > and < proofs, the `inputCoder` deals with the logic required to make a < proof work. i.e. for a < proof, the flow is that a user submits a negative kPublic and then the inputCoder will behind the scenes do `(constants.GROUP_MODULUS).add(kPublic)`. Interested in thoughts on this flow and whether submitting a negative kPublic, when trying to prove that `k_1 < kPublic` feels right

2) `(protocol/contracts/ACE/validators/publicRange/PublicRangeABIEncoder.sol)`: Currently, in the bytes proofOutput object I am returning a non-zero outputNotes field - I am returning the utility note. The alternative is to return an empty field, as this output/utility note will not be used for anything after proving the greater than/less than property. Interested in thoughts on this? Returning an output note is consistent with the other proofs and function setups

## Packages Updated
- [x] aztec.js - Javascript proof construction and verification code, ABI encoding public range proof input coder and tests
- [x] protocol - zero-knowledge proof validation smart contract, ABI encoding contract and tests

## Checklist
_aztec.js_
- [x] Javascript proof construction code
- [x] Tests for well formed proof data
- [x] Javascript proof verification code
- [x] Tests for successful validation and checking of failure states
- [x] ABI input encoding 
- [x] Tests for ABI input encoding

_protocol_
- [x] zk proof validator contract
- [x] Basic tests for zk proof validator
- [x] ABI output encoding contract
- [x] Basic tests for ABI output encoding contract
- [x] Extensive tests for zk validator
- [x] Tests for case where validator is checking a less than or equal to statement


packages/aztec.js/package.json
```diff
@@ -14,7 +14,7 @@
     "ethereumjs-abi": "^0.6.7",
     "ethereumjs-util": "^6.1.0",
     "web3-eth-abi": "1.0.0-beta.50",
-    "web3-utils": "1.0.0-beta.50"
+    "web3-utils": "1.0.0-beta.52"
```
PaulRBerg: Unless your other changes depend on this version of the utils, I would offload these updates onto @dependabot-bot.

packages/aztec.js/src/proof/publicRange/index.js
```diff
@@ -0,0 +1,180 @@
...
+};
+
+/**
+ * Encode a dividend computation transaction
+ *
+ * @method encodePublicRangeTransaction
+ * @memberof module:publicRange
+ * @param {Note[]} inputNotes input AZTEC notes
+ * @param {Note[]} outputNotes output AZTEC notes
+ * @param {Number} kPublic - public integer being compared against
```
ArnSch: I'm not a fan of reusing kPublic (which is an unclear variable name already in joinSplit). It seems like it will just add confusion as to what kPublic is. Can we not name it `publicComparaison` or something similar?

packages/aztec.js/src/proof/publicRange/index.js
```diff
@@ -0,0 +1,180 @@
...
+ * @param {string} senderAddress the Ethereum address sending the AZTEC transaction (not necessarily the note signer)
+ * @returns {Object} AZTEC proof data and expected output
+ */
+publicRange.encodePublicRangeTransaction = ({ inputNotes, outputNotes, kPublic, senderAddress }) => {
```
ArnSch: Why are we constraining ourselves to this really un-descriptive API? In this context, I wouldn't expect anyone to have any clue what inputNotes, outputNotes, or kPublic are supposed to be. 

Private range, for example, is much more descriptive. If we go in that direction, we can also feed in a boolean to denote whether this is a < or a > range comparison.

packages/aztec.js/src/proof/publicRange/index.js
```diff
@@ -0,0 +1,180 @@
...
+
+const publicRange = {};
+publicRange.verifier = verifier;
+
+/**
+ * Construct blinding factors
+ *
+ * @method constructBlindingFactors
+ * @param {Object[]} notes AZTEC notes
+ * @param {Number} kPublicBN - BN instance of the public integer being compared against
```
ArnSch: Misleading documentation

packages/aztec.js/test/abiEncoder/encoderFactory.js
```diff
@@ -453,4 +449,66 @@ describe('abiEncoder.encoderFactory', () => {
             expect(recoveredMetadata).to.equal(abiEncoder.encoderFactory.encodeMetadata(outputNote).data);
         });
     });
+
+    describe('Public range proof', () => {
+        it('should format publicRange properly', async () => {
+            // Setup
+            let accounts = [];
+            let notes = [];
+            const u = 10;
```
ArnSch: u?

packages/aztec.js/test/proof/publicRange/verifier.spec.js
```diff
@@ -0,0 +1,154 @@
...
+const publicRange = require('../../../src/proof/publicRange');
+const proofUtils = require('../../../src/proof/proofUtils');
+
+const { errorTypes } = utils.constants;
+
+const { expect } = chai;
+
+describe('Public range proof verifier', () => {
+    describe('Success States', () => {
+        it('should construct a valid dividend computation proof', async () => {
```
ArnSch: Invalid test name.

packages/protocol/contracts/ACE/validators/publicRange/PublicRange.sol
```diff
@@ -0,0 +1,325 @@
+pragma solidity >=0.5.0 <0.6.0;
+import "./PublicRangeABIEncoder.sol";
+
+/**
+ * @title Library to validate AZTEC Bilateral Swap zero-knowledge proofs
```
ArnSch: Please update the documentation.

packages/protocol/contracts/ACE/validators/publicRange/PublicRange.sol
```diff
@@ -0,0 +1,325 @@
...
+ * calls to OptimizedAZTEC can be treated as pure functions as this contract 
+ * contains no storage and makes no external calls (other than to precompiles)
+ * Copyright Spilbury Holdings Ltd 2019. All rights reserved.
+ **/
+contract PublicRange {
+
+    /**
+     * @dev PublicRange.sol will take any transaction sent to it and attempt to validate a zero knowledge proof.
+     * If the proof is not valid, the transaction throws.
+     * @notice See BilateralSwapInterface for how method calls should be constructed.
```
ArnSch: See above.

packages/protocol/contracts/ACE/validators/publicRange/PublicRange.sol
```diff
@@ -0,0 +1,325 @@
...
+     * AZTECBilateralSwap is written in YUL to enable manual memory management and for other efficiency savings.
+     **/
+    // solhint-disable payable-fallback
+    function() external {
+        assembly {
+
+            // We don't check for function signatures, there's only one function that 
+            // ever gets called: validateBilateralSwap()
+            // We still assume calldata is offset by 4 bytes so that we can represent 
+            // this contract through a comp\atible ABI
```
ArnSch: Typo

ArnSch (changes requested):
I don't think this is good to go, mostly because this code contains quite a few copy-paste errors, and because the API to proof construction for this proof isn't very good.

To answer your two questions specifically:

1) I think this whole interface need uphauling.
2) I lean slightly in the direction of not returning data just for the sake of it, but up to you.

packages/aztec.js/src/proof/publicRange/index.js
thomas-waite: Sure, done

packages/aztec.js/src/proof/publicRange/index.js
thomas-waite: Yep, fixed

packages/aztec.js/test/abiEncoder/encoderFactory.js
thomas-waite: Done

packages/aztec.js/test/proof/publicRange/verifier.spec.js
thomas-waite: Done

packages/protocol/contracts/ACE/validators/publicRange/PublicRange.sol
thomas-waite: Done

packages/protocol/contracts/ACE/validators/publicRange/PublicRange.sol
thomas-waite: Done

packages/protocol/contracts/ACE/validators/publicRange/PublicRange.sol
thomas-waite: Done

packages/aztec.js/src/proof/publicRange/index.js
thomas-waite: Agreed, changed the interface to:

```
publicRange.encodePublicRangeTransaction = async ({
    originalNote,
    publicComparison,
    senderAddress,
    isGreaterOrEqual,
    utilityNote,
}) => { }
```

The utilityNote is an optional parameter - the default is that it's generated internally and is not required as an input - but some flows require it to be possible to explicitly input

packages/aztec.js/package.json
thomas-waite: Good point

PR #91, opened 2019-04-23
feat: private range proof

thomas-waite:
## Summary
This PR implements a zero-knowledge proof that a confidential digital asset has a value greater or less than another confidential digital asset. This is a private range range proof.

This PR includes: JavaScript proof construction and verification code, ABI input encoding, smart contract zk proof validator, smart contract output ABI encoding and tests. 

## What I would like feedback on/where to focus
1) `(packages/aztec.js/src/proof/privateRange)': The API for `privateRange.encodePrivateRangeTransaction()` - it's inputs are formatted slightly differently to the other proofs. Specifically, rather than taking `inputNotes` and `outputNotes` as inputs, in the context of this proof I thought it would be more intuitive for the user to pass in `originalNote`, `comparisonNote` and `utilityNote`

2) `(packages/aztec.js/src/proof/privateRange)': I have not mixed `publicOwner`, `publicValue` into the hash used to construct the challenge (`proofUtils.constructChallenge()`) as they are constant for all possible instantations of this proof. Thoughts on this and other variables to mix in? (currently put senderAddress, notes and blindingFactors in)

3) `(packages/protocol/contracts/ACE/validators/privateRange/PrivateRange.sol)`: Due to the balancing relation equation in this proof, constructing a less than proof (rather than a greater than proof) I believe would require the user to create a greater than proof to the same effect
- e.g. if you want to prove that k_1 < k_2, then you would have to use this proof to show that k_2 > k_1. With my current proof relation, I don't believe I can directly show that k_1 < k_2, like you would be able to in a public range proof

4) `(packages/protocol/contracts/ACE/validators/privateRange/PrivateRangeABIEncoder.sol)`: Currently, in the `bytes proofOutput` object I am returning a non-zero `outputNotes` field - I am returning the utility note. The alternative is to return an empty field, as this output/utility note will not be used for anything after proving the greater than/less than property. Interested in thoughts on this? Returning an output note is consistent with the other proofs and function setups 

## Packages updated
- [x] aztec.js - Javascript proof construction and verification code, ABI encoding public range proof input coder and tests
- [x] protocol - zero-knowledge proof validation smart contract, ABI encoding contract and tests

## Checklist:

_aztec.js_
- [x] Javascript proof construction code
- [x] Tests for well formed proof data
- [x] Javascript proof verification code
- [x] Tests for successful validation and checking of failure states
- [x] ABI input encoding 
- [x] Tests for ABI input encoding

_protocol_
- [x] zk proof validator contract
- [x] Basic tests for zk proof validator
- [x] ABI output encoding contract
- [x] Basic tests for ABI output encoding contract
- [x] Extensive tests for zk validator


packages/aztec.js/src/proof/privateRange/index.js
```diff
@@ -0,0 +1,161 @@
...
+const { padLeft } = require('web3-utils');
+const crypto = require('crypto');
+
+const bn128 = require('../../bn128');
+const proofUtils = require('../proofUtils');
+const verifier = require('./verifier');
+const Keccak = require('../../keccak');
+
+const privateRange = {};
+privateRange.verifier = verifier;
```
zac-williamson, line 30: We should avoid assigning members of one module into members of another module, as it opens up the possibility of a code injection attack (if one can tamper with the client, you can swap this method out with a second method that, for example, has the same functionality but records every input parameter when this method is called)

packages/aztec.js/src/proof/privateRange/index.js
```diff
@@ -0,0 +1,161 @@
...
+ * @returns {string[]} proofData - constructed cryptographic proof data
+ * @returns {string} challenge - crypographic challenge variable, part of the sigma protocol
+ */
+privateRange.constructProof = (notes, sender) => {
+    // rolling hash is used to combine multiple bilinear pairing comparisons into a single comparison
+    const rollingHash = new Keccak();
+
+    proofUtils.parseInputs(notes, sender);
+
+    notes.forEach((note) => {
```
zac-williamson: Do we check that there are only 3 notes in this method?

packages/aztec.js/src/proof/privateRange/index.js
```diff
@@ -0,0 +1,161 @@
...
+
+        // Only set the first 2 values of kBar - the third is later inferred
+        // from a cryptographic relation. Set the third to 0
+        if (i <= 1) {
+            kBar = notes[i].k
+                .redMul(challenge)
+                .redAdd(blindingFactor.bk)
+                .fromRed();
+        } else {
+            kBar = padLeft(new BN(crypto.randomBytes(32), 16).umod(bn128.curve.n).toString(16), 64);
```
zac-williamson, line 120: Should this be 0? I don't follow why we set this to be a random variable

packages/aztec.js/src/proof/privateRange/index.js
```diff
@@ -0,0 +1,161 @@
...
+        }
+        const aBar = notes[i].a
+            .redMul(challenge)
+            .redAdd(blindingFactor.ba)
+            .fromRed();
+
+        return [
+            `0x${padLeft(kBar.toString(16), 64)}`,
```
zac-williamson, line 128: (this isn't really part of the review, but we should add a TODO task to perform this formatting in the abi encoder)

packages/aztec.js/src/proof/privateRange/index.js
```diff
@@ -0,0 +1,161 @@
...
+            `0x${padLeft(aBar.toString(16), 64)}`,
+            `0x${padLeft(notes[i].gamma.x.fromRed().toString(16), 64)}`,
+            `0x${padLeft(notes[i].gamma.y.fromRed().toString(16), 64)}`,
+            `0x${padLeft(notes[i].sigma.x.fromRed().toString(16), 64)}`,
+            `0x${padLeft(notes[i].sigma.y.fromRed().toString(16), 64)}`,
+        ];
+    });
+    return {
+        proofData,
+        challenge: `0x${padLeft(challenge.toString(16), 64)}`,
```
zac-williamson, line 139: Do we restrict challenge to be modulo the generator order? We should be validating this is the case in the smart contract

packages/aztec.js/src/proof/privateRange/index.js
```diff
@@ -0,0 +1,161 @@
...
+ * @method encodePrivateRangeTransaction
+ * @memberof module:privateRange
+ * @param {Note[]} originalNote original AZTEC note being to be compared
+ * @param {Note[]} comparisonNote note being compared against
+ * @param {Note[]} utilityNote additional note required to construct a valid balancing relationship
+ * @param {string} senderAddress the Ethereum address sending the AZTEC transaction (not necessarily the note signer)
+ * @returns {Object} AZTEC proof data and expected output
+ */
+privateRange.encodePrivateRangeTransaction = ({ originalNote, comparisonNote, utilityNote, senderAddress }) => {
+    const inputNotes = [...originalNote, ...comparisonNote];
```
zac-williamson: We should probably make the utility note inside `privateRange.constructProof`. This utility note should be created explicitly for this proof and then not re-used. If we allow the note to be passed as an input parameter it opens up the possibility that an implementor re-uses notes (and note secrets) and we want to avoid that possibility.  
  
It also makes the interface a bit simpler as the user has to keep track of fewer notes.

packages/aztec.js/src/proof/privateRange/index.js
```diff
@@ -0,0 +1,161 @@
...
+    proofUtils.parseInputs(notes, sender);
+
+    notes.forEach((note) => {
+        rollingHash.append(note.gamma);
+        rollingHash.append(note.sigma);
+    });
+
+    const blindingFactors = privateRange.constructBlindingFactors(notes, rollingHash);
+
+    const challenge = proofUtils.computeChallenge(sender, notes, blindingFactors);
```
zac-williamson: I think we should include `kPublic` and `publicOwner` in this challenge, even though they are 0. We want to have a consistent 'interface' for constructing AZTEC challenges, this will become more relevant when we start to implement virtual machine functionality

packages/aztec.js/src/proof/privateRange/verifier.js
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * @method verifyProof
+ * @memberof module:privateRange
+ * @param {Object[]} proofData - proofData array of AZTEC notes
+ * @param {string} challengeHex - challenge variable used in zero-knowledge protocol
+ * @param {string} sender - Ethereum address
+ * @returns {boolean} valid - boolean that describes whether the proof verification is valid
+ * @returns {option} errors - an array of all errors that were caught
+ */
+verifier.verifyProof = (proofData, challengeHex, sender) => {
```
zac-williamson, line 24: :thumbsup:

packages/aztec.js/src/proof/privateRange/index.js
```diff
@@ -0,0 +1,161 @@
+/**
+ * Constructs AZTEC private range zero-knowledge proofs
+ *
+ * @module privateRange
```
zac-williamson, line 4: We should probably add a bit of documentation here about how this works, specifically with regards to the input note and the comparison note. These notes need to be switched around, to convert this proof from a 'greater than or equals' proof to a 'less than or equals' proof, we should mention this

packages/protocol/contracts/ACE/validators/privateRange/PrivateRange.sol
```diff
@@ -0,0 +1,355 @@
...
+                let index := add(add(notes, mul(i, 0xc0)), 0x60)
+                calldatacopy(add(0x2c0, mul(i, 0x80)), index, 0x80)
+                }
+                mstore(0x00, keccak256(0x2c0, mul(n, 0x80)))
+            }
+        }
+    
+        // if we've reached here, we've validated the private range proof and haven't thrown an error.
+        // Encode the output according to the ACE standard and exit.
+        PrivateRangeABIEncoder.encodeAndExit();
```
zac-williamson, line 351: we should probably hoist this to the start of the fallback function. i.e. just after validatePrivateRange() is called. I know this isn't something we do in our other contracts, we should probably change all of them to make the program flow a bit clearer

packages/protocol/contracts/ACE/validators/privateRange/PrivateRange.sol
```diff
@@ -0,0 +1,355 @@
...
+                    } 
+                    case 0 {
+                        k := calldataload(noteIndex)
+                    }
+
+                    // Check this commitment is well formed...
+                    validateCommitment(noteIndex, k, a)
+
+                    // Set k = kx_j, a = ax_j, c = cx_j, where j = i - (m+1)
+                    if gt(i, 0x00) {
```
zac-williamson: (small optimiation, but we can replace `if gt(i, 0x00)` with `if i`. The `jumpi(x, location)` opcode will execute the jump instruction if `x` is anything but zero.

packages/protocol/contracts/ACE/validators/privateRange/PrivateRange.sol
```diff
@@ -0,0 +1,355 @@
...
+                        mstore(
+                            0x200,
+                            sub(0x30644e72e131a029b85045b68181585d97816a916871ca8d3c208c16d87cfd47, mload(0x100))
+                            )
+                    }
+
+                    // If i > m + 1 (i.e. subsequent output notes)
+                    // then we add \sigma^{-c} and \sigma_{acc} and store result at \sigma_{acc} (0x1e0:0x200)
+                    // we then calculate \gamma^{cx} and add into \gamma_{acc}
+                    if gt(i, m) {
```
zac-williamson: m = 2 and n = 3, so this never gets called.

We should discuss in person but I think we should include both all 3 notes in the pairing check. Depending on whether this is a 'greater than' or 'less than' proof, the input and comparison notes get swapped around.

I think we should always include the comparison note in the range proof check, as the contract calling this range proof check may not have any agency over the structure of this note (it could be prover defined). So we should include all 3 notes in the pairing check.

zac-williamson (changes requested):
Heya! Great job on this, I have a couple of comments surrounding the structure of the zero knowledge proof, but overall this is a solid implementation of our range proof.

packages/aztec.js/src/proof/privateRange/index.js
ArnSch: It's not clear to me that this is a problem introduced by this pattern. Even if the fn definition for `verifier` happened in the privateRange module, someone with the ability to tamper with the client could simply define a file which requires `privateRange` and redefines `privateRange.verifier`, and as long as that file is loaded prior to the actual use of `privateRange.verifier` in the client, the malicious, replaced fn will be called.

In fact, we use this pattern in tests to replace the method which gets a point from the trusted setup database.

If I misunderstood the attack you were thinking about, could you clarify? Otherwise I think duplicating the `verifier` code is not worth it.

packages/aztec.js/src/proof/privateRange/index.js
thomas-waite: Good point, added in a check

packages/aztec.js/src/proof/privateRange/index.js
thomas-waite: Agreed, added it to my TODO list

packages/aztec.js/src/proof/privateRange/index.js
thomas-waite: Done

packages/protocol/contracts/ACE/validators/privateRange/PrivateRange.sol
thomas-waite: Nice, done

packages/aztec.js/src/proof/privateRange/index.js
thomas-waite: Done

packages/aztec.js/src/proof/privateRange/index.js
thomas-waite: I agree about constructing the `utilityNote` internally to the proof construction. 

I have implemented, but have constructed inside `privateRange.encodePrivateRangeTransaction()` instead - I found it more straightforward to then perform the required tests. It also means that the `privateRange.constructProof()` method is solely focused on implementing the cryptography, rather than also constructing an additional note.

Constructed a `helper` function, with a `helpers.constructUtilityNote()` method to do this.

Interested in your thoughts

packages/protocol/contracts/ACE/validators/privateRange/PrivateRange.sol
thomas-waite: As discussed + reviewed, done

packages/aztec.js/src/proof/privateRange/index.js
thomas-waite: As discussed, leaving as is

packages/aztec.js/src/proof/privateRange/index.js
thomas-waite: When I load the the challenge variable from calldata in the validator contract, `PrivateRange.sol`, I load it under modular arithmetic by doing the following:

`let challenge := mod(calldataload(0x124), gen_order)`

calldata location 0x124 corresponds to the challenge. 

I think this covers it, but if not let me know

packages/aztec.js/src/proof/privateRange/index.js
thomas-waite: We currently throw an error during proof construction if the scalars are zero. So, setting kBar to 0 here would cause proof construction to fail. 

I agree that this behaviour is desired though, will look into it. Opened an issue to track this

packages/protocol/contracts/ACE/validators/privateRange/PrivateRange.sol
thomas-waite: Good idea, opening an issue to deal with this across all our validator contracts

PR #98, opened 2019-04-25
feat(contract-artifacts): make repo isomorphic

PaulRBerg:
## Description

The contract-artifacts is not isomorphic because it uses the `require-dir` package: https://github.com/aseemk/requireDir/issues/34

## Types of changes

## Checklist:


PaulRBerg:
TODO: find a better alternative to hardcoding the contract names in `contract-artifacts/src`. The beauty of require-dir is that you can just set-it-and-forget-it. That is, no matter what artifacts are being removed or added, it would find and export all of them.

PR #125, opened 2019-04-29
JoinSplit.sol sig validation to ZkAsset

thomas-waite:
## Summary
This PR moves ECDSA signature validation from the JoinSplit.sol validator contract, where it is currently performed, into the `ZkAsset.sol` `confidentialTransfer()` method.  

## Motivation
This change is motivated by a desire to create a distinction between the role of the validator contracts and the role of the `ZkAsset.sol` contract. The validator contracts determine the mathematical legitimacy of a transaction, whilst the `ZkAsset.sol` contract determines whether a transaction **should** happen. `ZkAsset.sol` performs the permissioning checks. 

## Where to focus/what I'd like feedback on
This PR includes a breaking change (described below), the `confidentialTransfer()` function now takes a second argument `bytes memory signatures`. 

This is necessary, as I have chosen not to encode the `inputSignatures` into the `proofData` (and therefore can't just supply the `proofData` and extract signatures inside `confidentialTransfer()`). 

I did this for gas optimisation reasons. The JoinSplit.sol validator no longer requires signatures, so sending them in the `proofData` would result in redundant information being sent. It costs **2,176 gas** to add 32 bytes of non-zero data to a transaction as input data. Signatures are 96 bytes long, so say the joinSplit has 2 input notes (many more are possible/likely) and this results in an additional **13,000 gas** cost.

Interested in your thoughts on the resulting breaking change and API change.

## Detailed changes
_aztec.js_
- Updated `inputCoder.js` and `encoderFactory.js` to remove `inputSignatures` from the ABI proof data encoding
- `inputSignatures` still generated inside the `encodeJoinSplitTransaction()` proof construction method, however adapted it to output the signatures as a third argument - rather than integrate into `proofData`. These signatures can then be passed when needed into the relevant smart contract method

_protocol_
- Removed signature validation from `JoinSplit.sol` and updated call data maps
- Performed EIP712 signature validation in the `confidentialTransfer()` method of `ZkAsset.sol`. This is the unilateral transfer method, for which only joinSplit proofs are possible
- Wrote `extractSignature()` and `validateSignature()` functions in `ZkAsset.sol` 
- Updated `confidentialTransfer()` method of `ZkAssetMintable.sol` to match new interface

## Types of changes
**Breaking change**
The interface for `confidentialTransfer()` in ZkAsset.sol has changed. It was previously:

 `function confidentialTransfer(bytes memory _proofData)`

and is now:
 `function confidentialTransfer(bytes memory _proofData, bytes memory _signatures)`

There is a second input argument, which takes an array of ECDSA signatures over all inputNotes which give permission for those notes to be used in a `confidentialTransfer()` operation. 

## Checklist:

_aztec.js_
-   [x] Remove `inputSignatures` from being in the joinSplit `proofData`, update `encoderFactory.js` and `InputCoder.js` to reflect this change
-   [x] Update `encodeJoinSplitTransaction()` to return `inputSignatures` as a third output - these can then be passed to the relevant smart contract method

_protocol_
-  [x] Remove signature validation from `JoinSplit.sol`, and update call data maps
-  [x] Add `bytes memory _signature` as a second input argument to the `ZkAsset.sol` method `confidentialTransfer()`
-  [x] Write `extractSignature()` function in ZkAsset.sol
-  [x] Write `validateSignature()` function in ZkAsset.sol, to validate an EIP712 signature in the ZkAsset domain 
-  [x] Perform signature validation in `confidentialTransfer()`, before get to `confidentialTransferInternal()`
-  [x] Update `ZkAssetMintable.sol` to use new `confidentialTransfer()` interface
-  [x] Write tests to check behaviour, specifically:
      - [x] JoinSplit.sol will perform validation with a fake signature (i.e. confirm it is no longer validating signatures)
      - [x] `confidentialTransfer()` in ZkAsset.sol will successfully validate a correct signature 
      - [x] `confidentialTransfer()` in ZkAsset.sol will fail for no signature 
      - [x] `confidentialTransfer()` in ZkAsset.sol will fail for a fake signature
      - [x] `confidentialTransfer()` in ZkAsset.sol will fail for an incorrect validator address (resulting in the EIP712 domain hash being incorrect)


packages/aztec.js/src/abiEncoder/inputCoder.js
```diff
@@ -32,6 +32,18 @@ inputCoder.bilateralSwap = (proofData, challenge, owners, metadata) => {
     return encoderFactory.encode(configs, abiParams, 'bilateralSwap');
 };
 
+/**
+ * Input encoding for burn proof data
+ *
+ * @method mint
```
ArnSch:
```suggestion
 * @method burn
```

ArnSch: (also, these docs are not very informative, but presume this is going to change a lot anyways through Paul's work)

packages/aztec.js/src/proof/joinSplit/index.js
```diff
@@ -219,33 +225,29 @@ joinSplit.encodeJoinSplitTransaction = ({
...
+                sender: senderAddress,
+            };
+            const { privateKey } = inputNoteOwners[index];
+            const { signature } = signer.signTypedData(domain, schema, message, privateKey);
+            const concatenatedSignature = signature[0].slice(2) + signature[1].slice(2) + signature[2].slice(2);
+            return concatenatedSignature;
+        });
+        signatures = `0x${signaturesArray.join('')}`;
+    } else {
+        signatures = [];
```
ArnSch: I think we want proof construction to exit if we have `inputNotes` but no `validatorAddress` or no `inputNoteOwners`, there is no point in proceeding and having the user have a hard to digest error about the proof not being valid if we can give more helpful feedback on client side.

This also simplifies the logic, since you can check if an error needs to be thrown, and if not just go through one code path (the map).

packages/protocol/contracts/ERC1724/ZkAsset.sol
```diff
@@ -131,49 +204,56 @@ contract ZkAsset is IZkAsset, IAZTEC, LibEIP712 {
...
-        logInputNotes(inputNotes);
-        logOutputNotes(outputNotes);
-
-        if (publicValue < 0) {
-            emit ConvertTokens(publicOwner, uint256(-publicValue));
-        }
-        if (publicValue > 0) {
-            emit RedeemTokens(publicOwner, uint256(publicValue));
+    /**
+    * @dev Perform ECDSA signature validation for a signature over an input note
```
ArnSch: Minor: not a great description of confidentialTransferInternal

packages/protocol/test/ACE/validators/joinSplit/signatureFlow.js
```diff
@@ -0,0 +1,195 @@
+/* global artifacts, expect, contract, beforeEach, it:true */
```
ArnSch: Can you double check that these tests abide by @PaulRBerg's new standards?

ArnSch: Specifically, I know Paul recently standardised how tests are named.

ArnSch (changes requested):
A couple things which are worth changing, but otherwise 👍

packages/aztec.js/src/abiEncoder/inputCoder.js
thomas-waite: Done, and yep I think so

packages/protocol/contracts/ERC1724/ZkAsset.sol
thomas-waite: Good point, done

packages/protocol/test/ACE/validators/joinSplit/signatureFlow.js
thomas-waite: Chatted with him, you're right. Updated test naming

packages/aztec.js/src/proof/joinSplit/index.js
thomas-waite: As discussed. 

Implementing a check for if `inputNoteOwners.length > 0 && validatorAddress.length === 0`, then throw. 

Otherwise, proceed. 

Replaced `inputNotes.map()` with `inputNoteOwners.map()`

PR #141, opened 2019-05-08
Local small point database for tests

thomas-waite:
## Summary
The elliptic curve point database is currently hosted remotely. This means that there has been a significant increase in the time taken to run tests - as each point for each test, used to construct a note, has to be fetched remotely. 

This PR adds a small point database locally into the `test` folders of the `aztec.js` and `protocol` packages, to decrease the time taken to perform tests. 

~14,000 points are added locally, representing ~1.4% of the original in house test trusted setup database. 

## Test speedups
% reduction in time taken to run package tests on my machine:
- **aztec.js** = 50% (24.5s -> 12.3s)
- **protocol** = 13% (437s -> 382s)

% increase in size after adding in the local setup database
- **aztec.js** = 3.7% (10.8Mb -> 11.2Mb)
- **protocol** = 50% (0.8Mb -> 1.2Mb)

## Types of changes
Changes focus around the `test` folders of `aztec.js` and `protocol`. 

There is now a small subset of the setup database and a `initDatabase.js' script. This initialisation script overrides the `setup.js` script in the source code, which fetches points remotely. 

In addition, the `initDatabase.js` script deals with functions elsewhere in the codebase that use K_MAX to instruct behaviour - for testing purposes there is now a smaller K_MAX, given the reduced database size. 

## Checklist:

_aztec.js_
-  [x] Add a small subsection of the point database locally to the `test` folder
-  [x] Write an initialisation script to adjust the default behaviour of fetching points remotely, and instead source them from the local database for testing purposes
-  [x] Manage functions throughout the codebase that use K_MAX - given the small local database, there is a new testing variable TEST_K_MAX
-  [x] Update tests

_protocol_
-  [x] Add a small subsection of the point database locally to the `test` folder
-  [x] Write an initialisation script to adjust the default behaviour of fetching points remotely, and instead source them from the local database for testing purposes
-  [x] Manage functions throughout the codebase that use K_MAX - given the small local database, there is a new testing variable TEST_K_MAX
-  [x] Update tests


packages/aztec.js/src/proof/proofUtils.js
```diff
@@ -6,7 +6,7 @@
 
 const {
     errors: { customError },
-    constants: { errorTypes, K_MAX },
+    constants,
```
ArnSch, line 9: We should be consistent on whether we destructure imports or not. In verifier.js for dividentComputation we breakout errorTypes, why not here?

packages/aztec.js/src/proof/proofUtils.js
```diff
@@ -201,12 +201,68 @@ proofUtils.convertTranscript = (proofData, m, challengeHex, errors, proofType) =
 };
 
 /**
- * Generate a random note value that is less than K_MAX
+ * Override the behaviour of this method in the src folder.
```
ArnSch: Can you add something about in which environments we override behaviour?

packages/aztec.js/test/initDatabase.js
```diff
@@ -0,0 +1,65 @@
+/**
+ * @module initDatabase
+ *
+ * This initialises the local trusted setup database, used for testing purposes.
+ * It contains 14,000 points - representing ~1.5% of the original in house trusted
```
ArnSch:
```suggestion
 * It contains 14,000 points - representing ~1.4% of the original in house trusted
```

packages/aztec.js/test/initDatabase.js
```diff
@@ -0,0 +1,65 @@
+/**
```
ArnSch, line 1: Can we add some printing when this file is loaded which says that we are doing this?

ArnSch (approved):
Looks great, some optional suggestions.

packages/aztec.js/test/initDatabase.js
thomas-waite: Good idea, done

packages/aztec.js/test/initDatabase.js
thomas-waite: Good catch, done

packages/aztec.js/src/proof/proofUtils.js
thomas-waite: Removed this comment that I added. It was left over from when I attempted something else and it doesn't add anything here. 

Should keep all local database initialisation comments/functionality to the `initDatabase.js` script

packages/aztec.js/src/proof/proofUtils.js
thomas-waite: I agree, we should standardise across the whole repo. This is a fairly large standardisation job though, and I think out of scope for this PR. 

I think Paul's refactor will go a long way toward standardising imports, so will leave this for now.

PR #168, opened 2019-05-15
fix: add dev tag to module docs to stop build fail

thomas-waite:
## Summary
Add the @dev tag to module comments for initDatabase.js

Docs was failing to build without it


ArnSch (approved):
LGTM

PR #170, opened 2019-05-16
Fix proof ids, POINTS_DB_URL, artifacts

thomas-waite:
## Problems
1) There was a redundant proofID defined in the `dev-utils` package - `UTILITY_PROOF`. It is not used for anything and was a placeholder left over from development. It has the same value as the `DIVIDEND_PROOF` ID, so was causing problems. 
2) The `POINTS_DB_URL` can not at the moment be updated. This is problematic if the URL does not behave as expected
3) The `PrivateRange.sol` was not deployed on testnets and no artifact was created for it

## Fixes
1) Removes the `UTILITY_PROOF` Id and updates the tests. Adds the `DIVIDEND_PROOF` Id into `IAZTEC`
2) Moves the `POINTS_DB_URL` definition into the `setup` module as a key:value pair - so it can be overwritten if need be
3) Write the `PrivateRange.sol` deployment script, and generate artifact

### DIFF
--- a/packages/aztec.js/src/setup/index.js
+++ b/packages/aztec.js/src/setup/index.js
@@ -15,9 +15,10 @@ const fetch = require('cross-fetch');
 const bn128 = require('../bn128');
 
 const compressionMask = new BN('8000000000000000000000000000000000000000000000000000000000000000', 16);
-const POINTS_DB_URL = 'https://ds8m7zxw3jpbz.cloudfront.net/data';
 
-const setup = {};
+const setup = {
+    POINTS_DB_URL: 'https://ds8m7zxw3jpbz.cloudfront.net/data',
+};
 
 /**
  * Decompress a 256-bit representation of a bn128 G1 element.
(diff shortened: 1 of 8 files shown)


packages/protocol/test/ACE/ACEAdjustSupply.js
```diff
@@ -667,25 +674,26 @@ contract('ACE Mint and Burn Functionality', (accounts) => {
             expect(burnReceipt.status).to.equal(true);
         });
 
-        it('should not update the validatedProofs mapping for utility proofs', async () => {
+        it.only('should not update the validatedProofs mapping for utility proofs', async () => {
```
ArnSch: ^

packages/aztec.js/src/setup/index.js
```diff
@@ -15,9 +15,10 @@ const fetch = require('cross-fetch');
 const bn128 = require('../bn128');
 
 const compressionMask = new BN('8000000000000000000000000000000000000000000000000000000000000000', 16);
-const POINTS_DB_URL = 'https://ds8m7zxw3jpbz.cloudfront.net/data';
 
-const setup = {};
+const setup = {
+    POINTS_DB_URL: 'https://ds8m7zxw3jpbz.cloudfront.net/data',
```
ArnSch, line 20: We should do a proper environment file if we're going to go for a dynamic URL.

ArnSch (changes requested):

packages/aztec.js/src/setup/index.js
thomas-waite: Don't have time to do this right now, but agree this is something that should be done

packages/protocol/test/ACE/ACEAdjustSupply.js
thomas-waite: Good spot, removed

PR #172, opened 2019-05-22
Fix zk assets

ArnSch:
**Note: This PR should not be merged unless automatic release to NPM is disabled.**

## Description

This pull request addresses two points:

1. Our zkAssetOwnable used to implement its own ownership logic, and now inherits from OpenZepplin's Ownable.sol
2. Our previous zkAsset contracts made it difficult for a user looking at deployed source code to tell whether the asset they were interacting with could adjust it's total supply. This PR makes this simpler by hardcoding the _canAdjustSupply flags based on the type of asset deployed, and by making the _canConvert be derived from the _linkedTokenAddress.

## Testing instructions

`npm run test`

I've added two test cases to zkAsset, both testing that the implicit setting of the canConvert flag given the value of _linkedTokenAddress behaves appropriately: `should set canConvert flag to false if 0x0 is linked token address` and `should set canConvert flag to true if linked token address is provided`

## Types of changes

* Breaking change (fix or feature that would cause existing functionality to change)


zac-williamson (approved):
👍

packages/protocol/migrations/4_deploy_zk_asset.js
```diff
@@ -26,9 +26,9 @@ module.exports = (deployer, network) => {
                 ZkAsset,
                 aceAddress,
                 erc20Address,
-                ERC20_SCALING_FACTOR,
-                canAdjustSupply,
-                canConvert
+                ERC20_SCALING_FACTOR
+                // canAdjustSupply,
```
thomas-waite: Think we can just remove // canAdjustSupply and // canConvert

PR #178, opened 2019-05-28
Fix confidentialTransferFrom() for ZkAssetMintable

thomas-waite:
**Note: This PR should not be merged into master unless automatic release to NPM is disabled.**

## Description
This PR solves a problem in that `confidentialTransferFrom()` for the `ZkAssetMintable.sol` contract did not mint tokens appropriately. 

If a user minted tokens using `confidentialMint()` and then converted/transferred them out using a joinSplit proof – it is possible that the tokens attempted to be transferred out are greater than the total number held by the contract, and so the call will revert.

To address this, logic has been added, similar to `confidentialTransfer()` that will mint the appropriate number of tokens from the linked ERC20 token to cover the shortfall. 

## Testing instructions
`npm run test`

I have added two success test cases to `ZkAssetMintable.js`. 

The first checks that when ACE has insufficient number of tokens, that the correct number are minted when `confidentialTransferFrom()` is called.

The second checks that ACE does not mint tokens unnecessarily when it has enough i.e. if publicValue <= totalSupply

## Types of changes

Bug fix - `confidentialTransferFrom() in `ZkAssetMintable.sol` did not correctly mint tokens.

## Notes
I am aware I have introduced some WET code in the redefining of `confidentialTransferFrom() in `ZkAssetMintable.sol`. 

Originally, I was going to pursue the approach taken in redefining `confidentialTransfer()` for `ZkAssetMintable` by defining a `confidentialTransferFromInternal()` in the parent `ZkAsset`.  However, realised that the proposed `confidentialTransferFromInternal()` and `confidentialTransferInternal()` would have the same code/perform the same function - and both could be combined into a more descriptive function such as `processTransaction()`. 

Will leave for a future refactor and focus this PR solely on the bug fix.

## Checklist:


packages/protocol/test/ZkAsset/ZkAssetMintable.js
```diff
@@ -15,13 +15,28 @@ const AdjustSupply = artifacts.require('./contracts/ACE/validators/AdjustSupply'
...
+const ZkAssetMintableTest = artifacts.require('./contracts/ZkAsset/ZkAssetMintableTest');
+
 
 const ZkAssetMintable = artifacts.require('./contracts/ZkAsset/ZkAssetMintable');
 
 
 AdjustSupply.abi = AdjustSupplyInterface.abi;
 JoinSplit.abi = JoinSplitInterface.abi;
 
+const signNote = (validatorAddress, noteHash, spender, privateKey) => {
```
ArnSch: Should this be in a util file?

packages/protocol/contracts/ZkAsset/ZkAssetMintableTest.sol
```diff
@@ -0,0 +1,27 @@
+pragma solidity >=0.5.0 <0.6.0;
```
ArnSch: I'm not a big fan of this pattern, especially not if it means having test contracts in a non-test directory. I understand that you're making sure that a contract can use confidentialTransferFrom, but I don't see what that teaches us compared to say having address[2] call these methods after being approved?

packages/protocol/test/ZkAsset/ZkAssetMintable.js
```diff
@@ -137,6 +152,208 @@ contract('ZkAssetMintable', (accounts) => {
...
+            // call confidentialApprove()
+            const indexes = [0, 1];
+            await Promise.all(indexes.map((i) => {
+                const { signature } = signNote(
+                    zkAssetMintable.address,
+                    mintedNotes[i].noteHash,
+                    zkAssetMintableTest.address,
+                    inputNoteOwners[i].privateKey
+                );
+                const concatenatedSignature = signature[0] + signature[1].slice(2) + signature[2].slice(2);
```
ArnSch: This is... ugly. I believe this got changed a lot on `develop` and cleaned away so I think all good.

ArnSch (approved):
Some minor points but otherwise LGTM

packages/protocol/test/ZkAsset/ZkAssetMintable.js
thomas-waite: Yep, we have since addressed this on the `develop` branch 

All signing related functions are now in a `signer` module

packages/protocol/test/ZkAsset/ZkAssetMintable.js
thomas-waite: I agree. It did get changed to an extent on `develop`, there is a function defined at the top of the `ZkAssetOwnable.js` script: 

    const confidentialApprove = async (indexes, notes, aztecAccounts) => {
        await Promise.all(
            indexes.map((i) => {
                const signature = signer.signNote(
                    zkAssetOwnable.address,
                    notes[i].noteHash,
                    zkAssetOwnableTest.address,
                    aztecAccounts[i].privateKey,
                );
                // eslint-disable-next-line no-await-in-loop
                return zkAssetOwnable.confidentialApprove(notes[i].noteHash, zkAssetOwnableTest.address, true, signature);
            }),
        );
    };

It's a bit nicer, but there's still scope to improve - the function should probably be moved to a `utils` type module, would make it clearer for external users too. Will leave this as is on `master` for now, but make a change on `develop` in my current PR.

packages/protocol/contracts/ZkAsset/ZkAssetMintableTest.sol
thomas-waite: Yeah good point, that's a nicer way to do it - will update the tests.

We clean away this pattern of having test contracts in non-test directories on `develop`

PR #183, opened 2019-06-06
Refactor monorepo

PaulRBerg:
The big refactor, finally done. Although this primarily entails changes made to the proof constructions and its associated tests in both aztec.js and protocol, a lot of changes have also been made to reduce the size of the repo (the number of lines of code has been cut in half).


packages/aztec.js/src/note/index.js
```diff
@@ -1,171 +1,168 @@
...
+            /**
+             * Value of note. BN instance in bn128 group's reduction context
+             * @member {BN}
+             */
+            this.k = null;
+            /**
+             * AZTEC commitment point \gamma, a bn128 group element
+             * @member {Point}
+             */
+            this.gamma = bn128.curve.decodePoint(publicKey.slice(2, 68), 'hex');
```
zac-williamson, line 58: It might be helpful to describe the structure of `publicKey` in the comments (e.g. 1st 2 characeters are `0x`, next 128 are two elliptic curve points that encrypt the note value, with the `ephemeral` key taking up the rest of the space, which is used by the note owner to decrypt their note.

packages/aztec.js/src/proof/dividend/index.js
```diff
@@ -0,0 +1,138 @@
...
+const bn128 = require('../../bn128');
+const { inputCoder, outputCoder } = require('../../encoder');
+const { Proof, ProofType } = require('../proof');
+const ProofUtils = require('../utils');
+
+const { AztecError } = errors;
+const { BN128_GROUP_REDUCTION } = constants;
+
+class DividendProof extends Proof {
+    constructor(notionalNote, residualNote, targetNote, sender, za, zb, metadata = [residualNote, targetNote]) {
```
zac-williamson, line 28: We should probably add comments describing what these input parameters are.  

`notionalNote` : the note that ones is computing a dividend of.  
`residualNote` : the dividend computation will have an integer rounding error, this is represented by the residual note (as having a public remainder would leak information)  
`targetNote`: the note that is being produced  
`za/zb` : public integers that define a ratio between notionalNote and targetNote.

It is also probably imporant to document that this proof assumes that it is in the prover's interest for `targetNote` to be as large as possible. What we are actually checking, is that `targetNote <= (zb / za) * notionalNote. So a prover can, if they desire, create a `targetNote` that is smaller than the note they are entitled to.  

If one wishes to use this proof to compute a debt, the roles of `targetNote` and `notionalNote` need to be swapped around, so that it is still in the prover's interest to minimize `targetNote`.

packages/aztec.js/src/proof/dividend/verifier.js
```diff
@@ -0,0 +1,86 @@
...
+                .add(bn128.h.mul(aBarX))
+                .add(sigma.mul(challengeX).neg());
+
+            if (B.isInfinity()) {
+                challengeResponse.appendBN(ZERO_BN);
+                challengeResponse.appendBN(ZERO_BN);
+                this.errors.push(errors.codes.BAD_BLINDING_FACTOR);
+            } else {
+                challengeResponse.appendPoint(B);
+                if (B.x.fromRed().eq(ZERO_BN_RED) && B.y.fromRed().eq(ZERO_BN)) {
```
zac-williamson: Both of these should be ZERO_BN

packages/aztec.js/src/proof/joinSplitFluid/index.js
```diff
@@ -0,0 +1,106 @@
...
+                publicValue: this.publicValue,
+                publicOwner: this.publicOwner,
+                challenge: this.challengeHex,
+            },
+            {
+                inputNotes: [],
+                outputNotes: this.outputNotes.slice(1),
+                publicValue: this.publicValue,
+                publicOwner: this.publicOwner,
+                challenge: keccak256(this.challengeHex), // TODO: figure out why this is the challenge hex and not the note the challenge itself
```
zac-williamson: We can remote this comment, `web3.keccak256` will treat the input as hex-encoded, if the string starts with `0x`

packages/aztec.js/src/proof/proof.js
```diff
@@ -0,0 +1,184 @@
...
+        this.m = inputNotes.length;
+        this.outputNotes = outputNotes;
+        this.notes = [...inputNotes, ...outputNotes];
+        this.sender = sender;
+        this.publicOwner = publicOwner;
+        this.metadata = metadata;
+        if (BN.isBN(publicValue)) {
+            this.publicValue = publicValue;
+        } else if (publicValue < 0) {
+            this.publicValue = bn128.curve.n.sub(new BN(-publicValue));
```
zac-williamson: This file uses `bn128.curve.n` where we should probably use `GROUP_MODULUS`

packages/aztec.js/src/proof/proof.js
```diff
@@ -0,0 +1,184 @@
...
+
+    get challengeHex() {
+        return `0x${padLeft(this.challenge.toString(16), 64)}`;
+    }
+
+    get eth() {
+        return {
+            // Ethereum will automatically prepend a bytes array with an evm word that
+            // represents the length of the bytes array
+            output: `0x${this.output.slice(0x40)}`,
```
zac-williamson, line 59: How come these are different by 2 characters? we should probably document why this is the case

packages/aztec.js/src/proof/proof.js
zac-williamson: Worth adding, arithmetic over BN numbers that use `GROUP_MODULUS` as their 'reduction context' is not compatible with BN numbers that use `bn128.curve.n`, despite them having the same value (BN.js checks that the reduction context is 'the same', but it does this via an equality check, so what's actually being checked is the memory location of the reduction context).  

i.e. we should pick either `GROUP_MODULUS` or `bn128.curve.n` and use that value consistently across aztec.js

zac-williamson (commented):
Heya! Amazing work on the refactor, the codebase is so much more coherent now. I've checked over all of the code that touches crypto, I have a couple of comments around how we document proof construction but other than that I have nothing else to add 👍

packages/aztec.js/src/proof/proof.js
PaulRBerg: I described this in a document I shared on Slack.

PaulRBerg:
Thank you @zac-williamson! I just pushed a commit with all the fixes for your comments.

PR #185, opened 2019-06-09
Add multisig

ArnSch:
## Description

In order to administrate ACE, we require a multisig contract, with a timelock. I use 0x's multisig contracts, upgraded for compiler >0.5.0, and add an owner contract which instantiates a new ACE.

All onlyOwner functions of ACE should only be callable through the multisig, and after a timelock save for invalidateProof, which might need to be done immediately. invalidateProof still requires m of n confirms.

## Testing instructions

yarn run test

## Types of changes


thomas-waite (approved):
Looks good to me

PR #193, opened 2019-06-29
Feat public range proof

thomas-waite:
**This PR is similar to a previous PR for a publicRange proof that was opened and partly reviewed (https://github.com/AztecProtocol/AZTEC/pull/89). This PR largely migrates the old public range proof to the style of the recently refactored codebase**

## Summary
In many use cases, it is important to be able to prove that a confidential digital asset has a value greater than or less than a public integer. This has to be performed in zero-knowledge - that is the value of the original confidential digital asset is never revealed. We call this proof a public range proof.

This PR implements such a proof - providing the Javascript proof construction code and the smart contract proof validator.

## Description
The proof construction code can be found in `aztec.js` in the `proof` module. This constructs the proof that an input note is greater or less than a `publicInteger` that is also supplied. 

The smart contract validator and abi encoder can be found in the `protocol` package, under `contracts/ACE/validators/publicRange`.

## Testing instructions
`yarn test`

## Types of changes
**New feature (non-breaking change which adds functionality**
This PR implements a new feature - a publicRange proof. There are no breaking changes.

**Test corrections**
Some proof validation tests have been corrected. Across all proofs, false positives were being returned for the failure tests checking failure when certain variables were not mixed into the challenge hash. This was due to the challenge hash being hashed > once.

## Notes
The JavaScript part of this PR is similar to a previous PR that was reviewed. The proof validation aspect of this PR is also similar, but has not been reviewed before. 

## Future work/next steps
The `utilityNote` should be abstracted away from the proof construction interface, it is currently exposed. It serves as a 'behind the scenses' helper note used to construct a proof statement. 

Abstracting the `utilityNote` away such that it would be constructed inside the proof, would naively require making the constructor asynchronous - will consider the best way to avoid this.

### DIFF
--- a/packages/aztec.js/src/proof/publicRange/helpers.js
+++ b/packages/aztec.js/src/proof/publicRange/helpers.js
@@ -39,7 +39,7 @@ helpers.constructUtilityNote = async (originalNote, publicInteger) => {
         }
         const utilityNote = await note.create(aztecAccount.publicKey, utilityValue);
         notes = [originalNote, utilityNote];
-    };
+    }
     return notes;
 };
 
--- a/packages/aztec.js/src/proof/publicRange/helpers.js
+++ b/packages/aztec.js/src/proof/publicRange/helpers.js
@@ -6,7 +6,7 @@ const helpers = {};
  * Check that the publicComparison integer is well formed. Specifically, check that it is positive and
  * that it is an integer.
  * @method checkPublicComparisonWellFormed
- * @param {Number} publicComparison - array of proof data from proof construction
+ * @param {Number} publicComparison - publicly visible integer, against which an AZTEC note is being compared
  */
 helpers.checkPublicComparisonWellFormed = (publicComparison) => {
     if (publicComparison < 0) {
@@ -28,7 +28,7 @@ helpers.checkPublicComparisonWellFormed = (publicComparison) => {
 /**
  * Check whether publicComparison is an integer
  *
- * @method isFloat
+ * @method isInteger
  * @param {Number} number - JavaScript number to be checked whether it is a float or not
  * @returns boolean - true if input is an integer, false if it is a float
  */
--- a/packages/aztec.js/src/proof/publicRange/helpers.js
+++ b/packages/aztec.js/src/proof/publicRange/helpers.js
@@ -1,0 +1,80 @@
+const secp256k1 = require('@aztec/secp256k1');
+const { constants, errors } = require('@aztec/dev-utils');
+const note = require('../../note');
+
+const helpers = {};
+/**
+ * Construct a utility note for the publicRange proof
+ *
+ * @dev Utility note is used to construct a balancing relationship
+ * for the sigma protocol to prove. Balancing relationship is:
+ *
+ * originalNoteValue = publicInteger + utilityNoteValue
+ *
+ * @method constructUtilityNote
+ * @memberof module:publicRange
+ * @param {Note[]} originalNote - note whose value is being compared to publicInteger
+ * @param {Integer} publicInteger - publicInteger which the originalNote is being compared against
+ * @returns {Note[]} array of AZTEC notes; specifically including originalNote,
+ * comparisonNote and utilityNote
+ */
+helpers.constructUtilityNote = async (originalNote, publicInteger) => {
+    let notes = [];
+    let utilityValue;
+
+    const aztecAccount = secp256k1.generateAccount();
+    if (originalNote.k !== undefined && publicInteger !== undefined) {
+        const originalValue = originalNote.k.toNumber();
+
+        if (originalValue > publicInteger) {
+            utilityValue = originalValue - publicInteger;
+        } else if (originalValue === publicInteger) {
+            utilityValue = 0;
+        } else {
+            throw errors.customError(constants.errorTypes.INCORRECT_NOTE_RELATION, {
+                message: 'notes supplied with publicInteger being less than originalValue',
+                originalValue,
+                publicInteger,
+            });
+        }
+        const utilityNote = await note.create(aztecAccount.publicKey, utilityValue);
+        notes = [originalNote, utilityNote];
+    };
+    return notes;
+};
+
+/**
+ * Check that the publicInteger integer is well formed. Specifically, check that it is positive and
+ * that it is an integer.
+ * @method checkpublicIntegerWellFormed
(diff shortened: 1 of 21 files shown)


packages/aztec.js/src/proof/publicRange/helpers.js
```diff
@@ -0,0 +1,39 @@
+const { constants, errors } = require('@aztec/dev-utils');
+
+const helpers = {};
+
+/**
+ * Check that the publicComparison integer is well formed. Specifically, check that it is positive and
+ * that it is an integer.
+ * @method checkPublicComparisonWellFormed
+ * @param {Number} publicComparison - array of proof data from proof construction
```
ArnSch: Doc string needs to be updated

packages/aztec.js/src/proof/publicRange/helpers.js
```diff
@@ -0,0 +1,39 @@
...
+            publicComparison,
+            type: typeof publicComparison,
+        });
+    }
+};
+
+/**
+ * Check whether publicComparison is an integer
+ *
+ * @method isFloat
```
ArnSch: docstring rename

packages/aztec.js/src/proof/publicRange/index.js
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * Constructs a public range proof that a note is greater than or equal to, or less than or
+     * equal to a public integer. Control of whether a > or < proof is constructed is determined
+     * by an input boolean 'isGreaterOrEqual'.
+     *
+     * The balancing relation being satisfied is:
+     * originalNoteValue = publicComparison + utilityNoteValue
+     *
+     * @param {Object} originalNote the note that a user is comparing against the publicComparison
+     * @param {Object} publicComparison publicly visible integer, which the note is being compared against
```
ArnSch: Is this expecting an integer or a (BN?) object?

ArnSch (approved):
Only very minor comments from me, looks good to go

packages/aztec.js/src/proof/publicRange/helpers.js
thomas-waite: done

packages/aztec.js/src/proof/publicRange/helpers.js
thomas-waite: done

packages/aztec.js/src/proof/publicRange/index.js
thomas-waite: good point, it's expecting an integer. Updated docstring

packages/aztec.js/src/proof/publicRange/index.js
```diff
@@ -0,0 +1,174 @@
...
+     * @param {Object} originalNote the note that a user is comparing against the publicComparison
+     * @param {Number} publicComparison publicly visible integer, which the note is being compared against
+     * @param {string} sender Ethereum address of the transaction sender
+     * @param {bool} isGreaterOrEqual modifier controlling whether this is a greater than, or less
+     * than proof. If true, it is a proof that originalNoteValue > publicComparison. If false, it is a
+     * proof that originalNoteValue < publicComparison
+     * @param {Note} utilityNote a helper note that is needed to satisfy a cryptographic balancing relation
+     */
+    constructor(originalNote, publicComparison, sender, isGreaterOrEqual, utilityNote) {
+        const publicValue = constants.ZERO_BN;
```
zac-williamson, line 33: We should probably check to see if the inputs satisfy the proof relation, and throw an error if this is not the case

packages/protocol/contracts/ACE/validators/publicRange/PublicRange.sol
```diff
@@ -0,0 +1,351 @@
...
+                    result := and(result, staticcall(gas, 6, 0x120, 0x80, 0x160, 0x40))
+
+                    // \gamma_i^{k}h^{a} and \sigma^{-c} in memory block 0x160:0x1e0
+                    // Store resulting point B at memory index b
+                    result := and(result, staticcall(gas, 6, 0x160, 0x80, b, 0x40))
+
+                    // If i > 0 (i.e. if it is the utility note)
+                    // then we add \sigma^{-c} and \sigma_{acc} and store result at \sigma_{acc} (0x1e0:0x200)
+                    // we then calculate \gamma^{cx} and add into \gamma_{acc}
+                    if gt(i, 0) { // m = 0
```
zac-williamson: it doesn't look like we're checking the range of the input note. We should probably do so, we can't assume that the input note has come from an AZTEC note registry

packages/protocol/test/ACE/validators/publicRange/index.js
```diff
@@ -0,0 +1,393 @@
...
+                proof.challenge = proof.challenge.add(bn128.groupModulus);
+                proof.constructOutputs(); // why do we have to do this?
+                const data = proof.encodeABI();
+                const result = await publicRangeValidator.validatePublicRange(data, sender, bn128.CRS);
+                expect(result).to.equal(proof.eth.outputs);
+            });
+        });
+
+        describe('Failure states', () => {
+            it('should fail if notes do NOT balance', async () => {
```
zac-williamson, line 84: We should check the failure states when the input or utility note is created over an incorrect trusted setup private key.

zac-williamson (dismissed):
It looks great, I like how extensive the tests are.  

I've added a couple of comments - I think it would be wise to add the input note into the pairing check, and ideally we want to also validate that our pairing check protects against rogue trusted setups

packages/aztec.js/src/proof/publicRange/index.js
thomas-waite: Good idea, done

packages/protocol/test/ACE/validators/publicRange/index.js
thomas-waite: Yep agreed. This needs doing for all proofs, assigned a seperate task to do this

packages/protocol/contracts/ACE/validators/publicRange/PublicRange.sol
thomas-waite: yeah good point. I have now removed the `if` statement and I roll the `gamma` and `sigma` coordinates of all notes into the accumulators. 

The pairing check is then called on the accumulators

PR #208, opened 2019-07-18
Feat private range proof

thomas-waite:
**This PR is largely similar to a previous PR that was reviewed and approved by Zac (https://github.com/AztecProtocol/AZTEC/pull/91) and merged. This PR has migrated the previously merged proof to the new construction.**

## Summary
This PR implements a zero-knowledge proof that a confidential digital asset has a value greater or less than another confidential digital asset. This is a private range range proof.

This PR includes: JavaScript proof construction and verification code, ABI input encoding, smart contract zk proof validator, smart contract output ABI encoding and tests. 

## Types of changes
**New feature**
This PR implements a new feature - a privateRange proof. There are no breaking changes.

## Future work/next steps
The `utilityNote` should be abstracted away from the proof construction interface, it is currently exposed. It serves as a 'behind the scenses' helper note used to construct a proof statement.

Abstracting the `utilityNote` away such that it would be constructed inside the proof, would naively require making the constructor asynchronous - will consider the best way to avoid this.


packages/aztec.js/test/proof/privateRange/index.js
```diff
@@ -0,0 +1,48 @@
...
+            const numNotes = 3;
+            expect(proof.data.length).to.equal(numNotes);
+            expect(proof.challengeHex.length).to.equal(66);
+            expect(proof.data[0].length).to.equal(6);
+            expect(proof.data[1].length).to.equal(6);
+            expect(proof.data[2].length).to.equal(6);
+        });
+    });
+
+    describe('Failure States', () => {
```
zac-williamson, line 39: We should test that the proof fails if the input note is not within the desired range.

This should fail in two locations - when constructing the proof, we should return an error state, and we should also test that verifying such a proof fails

zac-williamson: Oh, I was looking at the wrong file! We should still not construct a proof over bad inputs though, as this could cause misuse of our API

packages/protocol/contracts/ACE/validators/privateRange/PrivateRange.sol
```diff
@@ -0,0 +1,346 @@
...
+                    // then we add \sigma^{-c} and \sigma_{acc} and store result at \sigma_{acc} (0x1e0:0x200)
+                    // we then calculate \gamma^{cx} and add into \gamma_{acc}
+                    if gt(i, 0) {  // performing bilinear pairing comparison on all notes
+                        mstore(0x60, c)
+
+                        result := and(
+                            result,
+                            and(
+                                and(
+                                    staticcall(gas, 6, 0x1a0, 0x80, 0x1e0, 0x40),
```
zac-williamson: We have 3 notes in this proof, but we are only checking the pairing relationship on 2 notes

In this section, we need to update the memory location that we reference when we call the ecAdd precompile, so that we store the 3 notes linearly in memory

packages/protocol/test/ACE/validators/privateRange/index.js
```diff
@@ -0,0 +1,324 @@
...
+            proof.constructChallengeRecurse([proof.sender, proof.publicValue, proof.publicOwner, proof.blindingFactors]);
+            proof.challenge = proof.challengeHash.redKeccak();
+            const data = proof.encodeABI();
+            await truffleAssert.reverts(
+                privateRangeValidator.validatePrivateRange(data, sender, bn128.CRS),
+                truffleAssert.ErrorType.REVERT,
+            );
+        });
+
+        it('should fail if blindingFactors NOT integrated into challenge variable', async () => {
```
zac-williamson, line 310: I think the following test would be useful to add, to check that we're validating our bilinear pairing correctly:

Create a situation where the input/output/utility note values satisfy the private range proof, but the utility note is generated over a different setup key than the other notes (i.e. `y` is not the same, so the bilinear pairing test should fail). 

We should repeat the above test for permutations of all 3 notes (so one where the input note is over the wrong setup key, and one for the output note)

zac-williamson (changes requested):
This is great work. I like how clean the proof construction interface is and how methodical the tests are. The API has come a long way!

I couldn't find any problems with the cryptography, but I think there are a couple of extra tests that we would benefit from, to validate that our bilinear pairing check is preventing rogue commitments (created from a different trusted setup) from being added into the proof.

packages/protocol/test/ACE/validators/privateRange/index.js
thomas-waite: Very good point. As discussed, we want to implement this across all proofs. Will add the task to our scheduling tool and break it out seperately from this PR

packages/aztec.js/test/proof/privateRange/index.js
thomas-waite: Good point. Have implemented this check for the privateRange proof construction, but ideally we should also do this for all proofs

PR #209, opened 2019-07-22
Refactor EIP712 to be consistent with MetaMask

thomas-waite:
## Summary
This PR makes the AZTEC EIP712 implementation consistent with that of MetaMask. 

MetaMask uses the package `eth-sig-util` to perform signing, specifically using the function `signTypedData()` - https://github.com/MetaMask/eth-sig-util. This outputs an ECDSA signature, with the parameters being output in the order: `(r, s, v)`.

This order of parameters is different to the AZTEC signing functions, which output in order`(v, r, s)`. The EVM precompile that is then used to perform signature verification, ecrecover, expects params to be in order `(v, r, s)`. 

This PR updates the AZTEC signature verification contracts and signing functions, so that signatures generated with MetaMask (signing over AZTEC notes) can be verified by the AZTEC contracts.

## Detailed changes
The major changes relate to two packages: `aztec.js` and `protocol`.

**aztec.js**
- `signer` module: Where the note signing functions return ECDSA parameters, they have been adjusted to be in the order `(r, s, v)` rather than `(v, r, s)`
- `proof` module: the `joinSplit.constructSignatures()` code has been refactored and largely broken out into a seperate method that is now in the `signer` module

**protocol**
- `LibEIP712.sol` - `recoverSignature()` adapted to deal with params in `(r, s, v)` order and rearrange into correct form for ecrecover precompile
- `ZkAsset.sol` - `extractSignature()`adapted to deal with params in different order

## Testing instructions
`yarn test`

Various additional tests have been added

## Types of changes
**Breaking change**
There are a couple of breaking changes:
1) `signer.signNote()` has been renamed to `signer.signNoteForConfidentialApprove()` to better describe what this function should be used for and seperate it from the other signing functions
2) The ECDSA parameter output order for note signing functions in the `signer` module have been changed from `(v, r, s)` to `(r, s, v)`

### DIFF
--- a/packages/aztec.js/src/signer/index.js
+++ b/packages/aztec.js/src/signer/index.js
@@ -5,10 +5,10 @@
  * @module signer
  */
 
-const { constants } = require('@aztec/dev-utils');
+const { constants, proofs } = require('@aztec/dev-utils');
 const secp256k1 = require('@aztec/secp256k1');
 const typedData = require('@aztec/typed-data');
-const { randomHex } = require('web3-utils');
+const { randomHex, padRight } = require('web3-utils');
 
 const signer = {};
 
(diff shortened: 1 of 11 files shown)


packages/aztec.js/src/signer/index.js
```diff
@@ -62,7 +64,43 @@ signer.signNote = (verifyingContract, noteHash, spender, privateKey) => {
         status,
     };
     const { signature } = signer.signTypedData(domain, schema, message, privateKey);
-    return signature[0] + signature[1].slice(2) + signature[2].slice(2);
+    const r = signature[1];
+    const s = signature[2].slice(2);
+    const v = signature[0].slice(-2);
+    return r + s + v;
```
joeandrews: Lets just have one function that returns the signature structure that is compatible with metamask (r + s + v).

joeandrews: Ignore previous comments

joeandrews (changes requested):

packages/aztec.js/src/signer/index.js
thomas-waite: Good idea, done

joeandrews (approved):

ArnSch:
@thomas-waite Did you notice the failure during tests?

thomas-waite:
yeah, fixing it today. Some problem that came about as a result of rebasing

ArnSch (approved):
LGTM

PR #213, opened 2019-07-24
Feat upgradeable note registries

ArnSch:
## Description

In order to guarantee the correct implementation of any operation affecting the state of note registries within the AZTEC ecosystem, all of the data and behaviour relating to note registries is encapsulated in the AZTEC Cryptography Engine.

It is very likely that we will need to modify the behaviour of note registries in future, if we want AZTEC to keep evolving without requiring a Hardfork. For example, we might want to allow for storage of notes directly in smart contract storage instead of emitting as logs if storage costs are reduced, or we may need to update the schema of the proofOutputs instruction object if we introduce new crypto systems.

Given the immutability of ACE, we need to break out note registries into their own upgradeable modules.

This PR makes use of the unstructured storage pattern developed by OpenZeppelin to break out note registries into a storage contract and a Behaviour contract. Each Behaviour contract can change the implementation of methods, and introduce new data structures as long as it abides by a set minimum API to ensure compatibility with ACE.

Each Behaviour contract has an associated Factory, which manages that Behaviour's lifecycle including initial deployment and retirement. Factories need to be deployed and registered by the owner of ACE, and are associated to a unique ID made up of three uint8s packed into a uint24:

1. an Epoch
2. a CryptoSystem
3. an AssetType

Any ZkAsset can upgrade to any note registry as long as the epoch of the new registry is greater or equal to the existing registry, and the asset type is the same.

A new Epoch should be declared whenever the structure of storage gets amended (always addition).

## Testing instructions

`yarn test`

## Types of changes

* Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

-   [x] Update docs.
-   [x] Rebase
-   [ ] After rebase, update ZkAssets to include upgrade methods


packages/protocol/contracts/ACE/ACE.sol
```diff
@@ -88,26 +89,7 @@ contract ACE is IAZTEC, Ownable, NoteRegistry {
...
-        registry.confidentialTotalMinted = newTotalNoteHash;
-
-        // Dealing with minted notes
-        (,
-        bytes memory mintedNotes, // output notes
-        ,
-        ) = _proofOutputs.get(1).extractProofOutput();
-
-        updateOutputNotes(mintedNotes);
+        registry.mint(_proofOutputs);
```
thomas-waite, line 91: Good idea to abstract the mint and burn code away into the registries

packages/protocol/contracts/ACE/noteRegistry/Manager.sol
```diff
@@ -0,0 +1,334 @@
...
+            // 1. get the factoryAddress slot
+            // 2. add (epoch * 0x10000) to the slot
+            // 3. add (cryptoSystem * 0x100) to the slot
+            // 4. add (assetType) to the slot
+            // i.e. the range of storage pointers allocated to factoryAddress ranges from
+            // factoryAddress_slot to (0xffff * 0x10000 + 0xff * 0x100 + 0xff = factoryAddress_slot 0xffffffff)
+
+            // Conveniently, the multiplications we have to perform on epoch, cryptoSystem and assetType correspond
+            // to their byte positions in _factoryId.
+            // i.e. (epoch * 0x10000) = and(_factoryId, 0xffff0000)
```
thomas-waite: I know docs are to be worked on - this might be a place to look at. Think `0xffff0000` should be `0xff0000`

packages/protocol/contracts/ACE/noteRegistry/Manager.sol
```diff
@@ -0,0 +1,334 @@
...
+        ) = registry.getRegistry();
+        require(canConvert == true, "note registry does not have conversion rights");
+        require(canAdjustSupply == true, "note registry does not have mint and burn rights");
+        IERC20(linkedToken).transferFrom(msg.sender, address(this), _value.mul(scalingFactor));
+        registry.supplementTokens(_value);
+    }
+
+    /**
+    * @dev Query the ACE for a previously validated proof
+    * @notice This is a virtual function, that must be overwritten by the contract that inherits from NoteRegistr
```
thomas-waite: minor typo: `noteRegistr`

packages/protocol/contracts/ACE/noteRegistry/Manager.sol
```diff
@@ -0,0 +1,334 @@
...
+    *
+    * @param _proof - unique identifier for the proof in question and being validated
+    * @param _proofHash - keccak256 hash of a bytes proofOutput argument. Used to identify the proof in question
+    * @param _sender - address of the entity that originally validated the proof
+    * @return boolean - true if the proof has previously been validated, false if not
+    */
+    function validateProofByHash(uint24 _proof, bytes32 _proofHash, address _sender) public view returns (bool);
+
+    function createNoteRegistry(
```
thomas-waite: If I understand it correctly, this `createNoteRegistry()` function here is creating a 'default' note registry - one in which the `defaultCryptoSystem` is being used. 

Is there scope to rename this method to `createDefaultNoteRegistry()`  to seperate it from the second `createNoteRegistry()` function? The second, I believe, gives the user agency over the type of `noteRegistry` they are creating

packages/protocol/contracts/ACE/noteRegistry/epochs/201907/Behaviour.sol
```diff
@@ -0,0 +1,345 @@
...
+import "../../../../interfaces/IAZTEC.sol";
+import "../../../../libs/NoteUtils.sol";
+import "../../interfaces/Behaviour.sol";
+
+/**
+ * @title NoteRegistryBehaviour contract which contains the storage variables that define the set of valid
+ * AZTEC notes for a particular address
+ * @author AZTEC
+ * @dev TODO
+ * Copyright Spilbury Holdings Ltd 2019. All rights reserved.
```
thomas-waite: minor typo: `Spilbury`

packages/protocol/test/ACE/NoteRegistry/NoteRegistryManager.js
```diff
@@ -0,0 +1,191 @@
...
+/* eslint-disable-next-line object-curly-newline */
+const { proofs } = require('@aztec/dev-utils');
+
+// ### Artifacts
+const ACE = artifacts.require('./ACE');
+const NoteRegistryFactory = artifacts.require('./noteRegistry/epochs/201907/convertible/FactoryConvertible201907');
+const Behaviour = artifacts.require('./noteRegistry/interfaces/NoteRegistryBehaviour');
+const ERC20Mintable = artifacts.require('./ERC20Mintable');
+
+const { BOGUS_PROOF, JOIN_SPLIT_PROOF } = proofs;
```
thomas-waite: BOGUS_PROOF and JOIN_SPLIT_PROOF are unusued, can delete

packages/protocol/test/ACE/NoteRegistry/NoteRegistryManager.js
```diff
@@ -0,0 +1,191 @@
...
+        const assetType = 0b01; // (adjust, canConvert) in binary;
+        await ace.setFactory(generateFactoryId(epoch, cryptoSystem, assetType), factoryContract.address, { from: owner });
+        erc20 = await ERC20Mintable.new();
+    });
+
+    describe('Success States', async () => {
+        it('should register a factory', async () => {
+            const receipt = await ace.setFactory(generateFactoryId(1, 2, 1), factoryContract.address, { from: owner });
+            const event = receipt.logs.find(l => l.event === 'SetFactory');
+            expect(event.args.epoch.toNumber()).to.equal(1);
```
thomas-waite, line 42: I've started using a useful external package, `truffleAssert`: https://kalis.me/check-events-solidity-smart-contract-test-truffle/, to abstract away some of the details involved in event testing

packages/protocol/contracts/ACE/noteRegistry/epochs/201907/Behaviour.sol
```diff
@@ -0,0 +1,345 @@
...
+
+    /**
+        * @dev Update the state of the note registry according to transfer instructions issued by a
+        * zero-knowledge proof
+        *
+        * @param _proof - unique identifier for a proof
+        * @param _proofOutput - transfer instructions issued by a zero-knowledge proof
+    */
+    function updateNoteRegistry(
+        uint24 _proof,
```
thomas-waite: unused variable, `uint24 _proof` - same for all Behaviour contracts in the `updateNoteRegistry()` functions

packages/protocol/contracts/ACE/noteRegistry/epochs/201907/adjustable/Behaviour.sol
```diff
@@ -0,0 +1,106 @@
...
+        ,
+        int256 publicValue) = _proofOutput.extractProofOutput();
+
+        updateInputNotes(inputNotes);
+        updateOutputNotes(outputNotes);
+
+        // If publicValue != 0, enact a token transfer
+        // (publicValue < 0) => transfer from publicOwner to ACE
+        // (publicValue > 0) => transfer from ACE to publicOwner
+        if (publicValue != 0) {
```
thomas-waite: This asset type is `adjustable`, not `mixed`. So, it shouldn't be able to convert `publicValue` into note form, or convert notes into `publicValue`. 

Therefore, this check should be removed as it implies that there's a possibility of `registry.canConvert` returning `true`. The `canConvert` flag is hardcoded to be `false`.

Input arg `uint24 _proof` is unused

packages/protocol/contracts/ACE/noteRegistry/epochs/201907/convertible/Behaviour.sol
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+    * @dev This should be called from an asset contract.
+    */
+    function publicApprove(address _publicOwner, bytes32 _proofHash, uint256 _value) public onlyOwner {
+        registry.publicApprovals[_publicOwner][_proofHash] = _value;
+    }
+
+    function setConfidentialTotalMinted(bytes32 newTotalNoteHash) internal onlyOwner returns (bytes32) {
```
thomas-waite: This Behaviour contract defines the behaviour for a `convertible` asset/note registry, so there's no concept of minting/burning. Can remove methods: `setConfidentialTotalMinted()` and `setConfidentialTotalBurned()`.

packages/protocol/contracts/ACE/noteRegistry/epochs/201907/mixed/Behaviour.sol
```diff
@@ -0,0 +1,135 @@
...
+        address publicOwner,
+        int256 publicValue) = _proofOutput.extractProofOutput();
+
+        updateInputNotes(inputNotes);
+        updateOutputNotes(outputNotes);
+
+        // If publicValue != 0, enact a token transfer
+        // (publicValue < 0) => transfer from publicOwner to ACE
+        // (publicValue > 0) => transfer from ACE to publicOwner
+        if (publicValue != 0) {
```
thomas-waite: This is my fault, I wrote the code that deals with transferring `publicValue` in the `updateNoteRegistry()` function, but line 93 to line 116 could probably be broken out into a seperate function: `convertPublicValue()` or similar. 

Might make what's happening more descriptive/clearer

packages/protocol/migrations/7_deploy_factory.js
```diff
@@ -0,0 +1,30 @@
...
+    if (isUndefined(ACE) || isUndefined(ACE.address)) {
+        console.log('Please deploy the ACE contract first');
+        process.exit(1);
+    }
+
+    /* eslint-disable no-new */
+    new Promise(() => {
+        return deployer.deploy(ConvertibleFactory, ACE.address).then(async ({ address }) => {
+            const ace = await ACE.at(ACE.address);
+            // const adjustableFactoryAddress = await AdjustableFactory.new(ACE.address);
```
thomas-waite: As you mentioned the docs are to be worked on next. Some comments and unused variables can be removed here and the script could perhaps be renamed to `7_deploy_convertibleFactory.js` to reflect the type of factory being deployed

packages/protocol/contracts/ACE/noteRegistry/proxies/AdminUpgradeabilityProxy.sol
```diff
@@ -0,0 +1,24 @@
...
+   * Contract constructor.
+   * @param _logic address of the initial implementation.
+   * @param _admin Address of the proxy administrator.
+   * @param _data Data to send as msg.data to the implementation to initialize the proxied contract.
+   * It should include the signature and the parameters of the function to be called, as described in
+   * https://solidity.readthedocs.io/en/v0.4.24/abi-spec.html#function-selector-and-argument-encoding.
+   * This parameter is optional, if no data is given the initialization call to proxied contract will be skipped.
+   */
+  constructor(address _logic, address _admin, bytes memory _data) UpgradeabilityProxy(_logic, _data) public payable {
+    assert(ADMIN_SLOT == bytes32(uint256(keccak256('eip1967.proxy.admin')) - 1));
```
thomas-waite: Out of interest, when we are generating a location for the storage slots by hashing a string - why do we subtract 1 off the result? 

Presume it's just to increase the randomness of the resulting storage slot and make a collision even less likely?

packages/protocol/test/ACE/NoteRegistry/epochs/201907/Proxy.js
```diff
@@ -0,0 +1,56 @@
...
+const truffleAssert = require('truffle-assertions');
+
+// ### Artifacts
+const AdminUpgradeabilityProxy = artifacts.require('./noteRegistry/proxies/AdminUpgradeabilityProxy.sol');
+const Behaviour = artifacts.require('./noteRegistry/epochs/201907/Behaviour201907.sol');
+
+
+const { BOGUS_PROOF, JOIN_SPLIT_PROOF } = proofs;
+
+contract('Proxy', (accounts) => {
```
thomas-waite, line 17: We should make time to implement more tests. Some initial ideas:

**Success states**
- Can upgrade the behaviour contract and link it to the proxy

**Failure states**
- Should fail to upgrade if not admin
- Should fail if low level `proxy.delegate()` function called on a non-existant contract. We want to check this bug that Trail of Bits discuss: https://blog.trailofbits.com/2018/09/05/contract-upgrade-anti-patterns/

packages/protocol/contracts/ACE/noteRegistry/interfaces/Behaviour.sol
```diff
@@ -0,0 +1,83 @@
...
+import "../Manager.sol";
+
+/**
+ * @title NoteRegistryBehaviour contract which contains the storage variables that define the set of valid
+ * AZTEC notes for a particular address
+ * @author AZTEC
+ * @dev TODO
+ * Copyright Spilbury Holdings Ltd 2019. All rights reserved.
+ **/
+contract NoteRegistryBehaviour is Ownable, IAZTEC {
```
thomas-waite: Is it worthwhile renaming this file from `Behaviour.sol` to `NoteRegistryBehaviour.sol`, to match the Solidity name of this contract/interface? That way when it's imported elsewhere, inherited and referred to as `NoteRegistryBehaviour` rather than `Behaviour` it's a bit clearer as to what contract we're referring to

packages/protocol/contracts/ACE/noteRegistry/interfaces/Behaviour.sol
```diff
@@ -0,0 +1,83 @@
...
+
+    bool public isActiveBehaviour;
+    bool public initialised;
+    address public dataLocation;
+
+    constructor () Ownable() public {
+        isActiveBehaviour = true;
+    }
+
+    function initialise(
```
thomas-waite: Given that this contract is just an interface, should the code for this `initialise()` function not be moved out? I would put it in `noteRegistry/epochs/201907/Behaviour.sol`, the base behaviour contract from which all other behaviour contracts inherit. 

Then for the interface, just put `function initialise(address _newOwner, etc) public`

thomas-waite (changes requested):
I really like this upgradeability work and the use of the unstructured storage upgradeability pattern. I don't have any high level design comments, my comments revolve around implementation. 

From a testing point of view, as discussed, we should make time for more proxy based tests. I've listed a few that might be good for us to implement. As you've pointed out, we should also implement `ZkAsset` tests to confirm that this upgradeability behaviour works with our current setup.

packages/protocol/contracts/ACE/noteRegistry/Manager.sol
ArnSch: I decided against this so that the interface the current ZkAssets use doesn't change. Both methods are still available to users, depending on arguments passed.

packages/protocol/contracts/ACE/noteRegistry/epochs/201907/Behaviour.sol
ArnSch: Purposeful, and should be documented: In future, we will need to be able to assert that a proof is supported by a particular note registry (for example, if we switch to a different storage pattern for the notes, proof outputs will need to change, so we will need a new generation of proofs).

Will add docs.

packages/protocol/contracts/ACE/noteRegistry/epochs/201907/convertible/Behaviour.sol
ArnSch: Will throw an "abstract contract" error, since these methods are on the base note registry interface (and need to be)

packages/protocol/contracts/ACE/noteRegistry/proxies/AdminUpgradeabilityProxy.sol
ArnSch: Detailed in the EIP 1967 spec (https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1967.md), the -1 is to reduce the chance of a pre-image attack

packages/protocol/contracts/ACE/noteRegistry/interfaces/Behaviour.sol
ArnSch: I'm not against this, but thought I'd rather always guarantee that this initialise function, even if not re-implemented, will at least transferOwnership and set the initialised flag.

It arguably does make it harder to notice that you are not setting your initial NoteRegistry state for example, but I'd expect that to be discovered through tests you would necessarily be writing, whilst the initialised flag is more unlikely to be caught.

I might add the ownership transfer test + initialised flag test to programatic tests and move this logic.

packages/protocol/contracts/ACE/noteRegistry/epochs/201907/adjustable/Behaviour.sol
ArnSch: Ditto

ArnSch: On input arg, see above comment.

This is required, since you need to explicitly fail if a valid proof has been sent with a kPublic !== 0, otherwise you will end up modifying noteRegistry state and potentially creating value out of thin air.

packages/protocol/contracts/ACE/noteRegistry/epochs/201907/Behaviour.sol
thomas-waite: Makes sense

packages/protocol/contracts/ACE/noteRegistry/proxies/AdminUpgradeabilityProxy.sol
thomas-waite: Makes sense

packages/protocol/contracts/ACE/noteRegistry/epochs/201907/mixed/Behaviour.sol
ArnSch: I think this is actually fine as is.

packages/protocol/contracts/ACE/noteRegistry/interfaces/Behaviour.sol
thomas-waite: Hmm, after reading your explanation I understand why it's like that - we must gurantee sure that the `initialise` method is available when a new behaviour is deployed and that it's not possible to initialise more than once. 

Perhaps, we could have a `baseBehaviour.sol` contract that includes the code for these important methods. The `epochs/201907/Behaviour.sol` could then inherit from `baseBehaviour.sol` rather than the interface. 

The interface would then just be a standard interface without the code in it

packages/protocol/contracts/ACE/noteRegistry/Manager.sol
thomas-waite: Fair enough

packages/protocol/contracts/ACE/noteRegistry/epochs/201907/adjustable/Behaviour.sol
thomas-waite: Yeah you're right

packages/protocol/test/ACE/NoteRegistry/epochs/201907/Proxy.js
ArnSch: Ended up adding `should delegate non-admin call even if fn name exists in proxy`, `should fail to upgrade if not admin`, `should fail to set implementation to 0x0`. The first success case you mentioned is tested in Manager.sol.

thomas-waite (approved):

PR #237, opened 2019-08-07
fix(protocol): wrong if statement in recoverSignature

LeilaWang:
Should enter the if statement when result is true but signer address is 0


ArnSch (approved):
LGTM, thanks for catching!

PR #244, opened 2019-08-11
Refactor meta data for granting view key access

thomas-waite:
## Summary
This PR refactors the treatment of `metaData`. `metaData` is intended to be a general purpose data field in an AZTEC note, to be filled with implementation specific data. It is not used in proof construction/verification. 

This PR aims to make it more general purpose, editable and easier to interact with. The changes have been made with specific attention to the use case of using it to store IES encrypted viewing keys. 
 

## Detailed changes
Changes focus around:
- note class schema and associated methods
- input coder
- output coder
- proof construction
- ZkAsset

### Note
The `note` schema has changed - it now includes an explicit `metaData` field. Previously, the field 'ephemeral' acted as the pseudo `metaData` field as until now that was intended to be the sole use. 

To support this additional property of the `note` class, two extra methods have been added to the `note` class:
1) _setMetaData()_ - allows a user to set the `metaData` of the note to a custom string
2) _exportMetaData()_ - exports the note `metaData` in an ABI compatible format, used by the `inputCoder`

The note property `ephemeral` has been removed, given that it is now located within the `metaData`.

### Input coder
There is now a refactored `encodeMetaData()` function in the `inputCoder` module. It has been made more general, such that it can handle dynamic `metaData` lengths rather than the previous static treatment. This is essential if the `metaData` is to be used to handle arbitrary length arrays of encrypted viewing keys. 

### Output coder
Given the note schema change, several of the `outputCoder` functions had to be updated. Notes are no longer of fixed size, as the refactored `metaData` is now of arbitrary size, and so several of the encoding functions in the `outputCoder` have had to be refactored to work with dynamic note and note data lengths.

In addition, `inputNotes` and `outputNotes` are treated differently with respect to `metaData` - `outputNotes` need their `metaData` encoding whilst `inputNotes` do not. So some functions that apply to both types of notes have been split out and refactored.

### Proof construction
A breaking change has been made to all proof construction interfaces. Given that the `note` schema has now been adapted to include the `metaData` of the note, there is now no need to pass the `metaData` of the `outputNotes` into proof construction. As a result, the `metaData` input parameter has been removed from all proofs. It is instead extracted from the notes internal to the proof. 

This also removes potential confusion as to why `metaData` is being passed to the proof construction methods. 

### ZkAsset
A method has been added `updateNoteMetaData()`.

This is a use case specific change. It is intended to allow a `noteOwner` or an address on a pre-approved list of addresses to update the `metaData` of a note. Changes are affected by the method emitting an event which includes the updated `metaData` of the note. 

When a user updates the note `metaData`, it is intended that they are adding another encrypted viewing key to it and so granting note viewing permission of their note to another user.

## Help wanted
I think there may be a bug in the swap ABI encoder contract. The last word of data it is outputting is the `metaData` of the second `outputNote`. I believe this should be the `metaData` of the second `inputNote`, this is what `proof.eth.outputs` returns. The input/output notes are reversed in a `swap`. 

## Checklist:

-   [x] Add to `note` class - property + methods
-   [x] Make outputCoder ntoe methods dynamic
-   [x] Implement `inputCoder.encodeMetaData()` method
-   [x] Edit proof construction interface
-   [x] Migrate all existing proof tests to adjusted interface
-   [x] Fix bug in `swapValidator` contract - it's placing the `metaData` for the second `outputNote` in the position of the `metaData` for the second `inputNote`
-   [x] Implement the approved address permissioning flow for `ZkAsset.updateNoteMetaData()` method access
-   [x] Remove the `ephemeral` field from the note schema
-   [x] Integration test with the extension work
-   [x] Fix minting problem where `totalMintedNote` does not contain the custom metaData


packages/protocol/contracts/ERC1724/base/ZkAssetBase.sol
```diff
@@ -291,17 +293,23 @@ contract ZkAssetBase is IZkAsset, IAZTEC, LibEIP712 {
...
 
-        // There should be a permission lock here requiring that only the noteOwner can call
-        // this function. It has been deliberately removed on a short term basis
-        emit UpdateNoteMetadata(noteOwner, noteHash, metadata);
+        require(
+            noteAccess[msg.sender] == noteHash || noteOwner == msg.sender,
+            'caller does not have permission to update metaData'
+        );
+
+        address addressToApprove = MetaDataUtils.extractAddresses(metaData);
```
joeandrews, line 303: Joe to write spec to make this work for actual meta data structure.

packages/aztec.js/src/encoder/outputCoder.js
```diff
@@ -11,39 +11,56 @@ const { keccak256, padLeft } = require('web3-utils');
...
         ephemeral = secp256k1.decompressHex(note.slice(0x1c0, 0x202));
     } else {
-        // inputNote
-        expectedLength = 0xc0;
-    }
-
-    if (length !== expectedLength) {
-        throw new Error(`unexpected note length of ${length}`);
+        // output note with custom metaData set, expectedLength is (0x20 * 4 + 0x20 * 2 + noteDataLength).toString(16);
+        ephemeral = secp256k1.decompressHex(note.slice(0x1c0, 0x202));
```
ArnSch, line 32: This if/else seems superfluous?

ArnSch: The only change between input note and output notes is the extraction of the ephemeral key and/or any additional metadata? If so, seems like you can simplify the API by simply always checking whether the length of the note contains/doesn't contain metadata.

packages/aztec.js/src/encoder/outputCoder.js
```diff
@@ -64,16 +81,26 @@ outputCoder.decodeNote = (note) => {
  *
  * @method decodeNotes
  * @param {note} notes - array of AZTEC notes
+ * @param {bool} isOutputNote - boolean determining whether the supplied notes are outputNotes.
+ * If they are, decodeOutputNote() is called rather than decodeInputNote()
  * @returns {Object[]} array of note variables - array of decoded and extracted note variables
  * where each element corresponds to the note variables for an individual note
  */
-outputCoder.decodeNotes = (notes) => {
+outputCoder.decodeNotes = (notes, isOutputNote) => {
```
ArnSch: This random bool feels like it shouldn't be there, we have enough information in the note to derive it without require the developer to input it.

packages/aztec.js/src/encoder/outputCoder.js
```diff
@@ -197,33 +224,19 @@ outputCoder.encodeNotes = (notes, isOutput) => {
...
-
-        const metaDataSize = 0x137;
-        noteDataLength = (0x20 * 2 + isMetadataPresent * metaDataSize).toString(16);
-        noteLength = (0x20 * 4 + 0x20 * 2 + isMetadataPresent * metaDataSize).toString(16);
-    } else {
-        encoded[7] = secp256k1.compress(note.ephemeral.getPublic()).slice(2);
-        noteLength = 'e1';
-        noteDataLength = '61';
-    }
+    const metaDataSize = parseInt(note.metaData.slice(2).length / 2, 10);
```
ArnSch: The encoding functions suffer from the same thing as the decoding functions: the only difference seems to be that some computed values need to take into account the length of the metadata. This is something we can derive internally to the function (with a simple if(note.metadata)).

packages/aztec.js/src/proof/joinSplitFluid/index.js
```diff
@@ -70,13 +62,13 @@ class JoinSplitFluidProof extends JoinSplitProof {
...
                         ...this.outputNotes[0],
-                        forceMetadata: true,
+                        switchEncoding: false,
                     },
                 ],
                 outputNotes: [
                     {
                         ...this.inputNotes[0],
-                        forceNoMetadata: true,
+                        switchEncoding: false,
```
ArnSch: I believe if you implement the switch proposed in the outputCoder module, you can get rid of these magic flags.

packages/aztec.js/test/encoder/inputCoder.js
```diff
@@ -0,0 +1,90 @@
...
+describe('Input coder', () => {
+    const account = secp256k1.generateAccount();
+    const noteValue = 10;
+    describe('encodeMetaData()', async () => {
+        it('should encode the metaData of multiple notes, when customData set', async () => {
+            const numNotes = 2;
+            const testNoteA = await note.create(account.publicKey, noteValue);
+            const testNoteB = await note.create(account.publicKey, noteValue);
+            const customData =
+                '0x00000000000000000000000000000028000000000000000000000000000001a4000000000000000000000000000000003339c3c842732f4daacf12aed335661cf4eab66b9db634426a9b63244634d33a2590f06a5ede877e0f2c671075b1aa828a31cbae7462c581c5080390c96159d5c55fdee69634a22c7b9c6d5bc5aad15459282d9277bbd68a88b19857523657a958e1425ff7f315bbe373d3287805ed2a597c3ffab3e8767f9534d8637e793844c13b8c20a574c60e9c4831942b031d2b11a5af633f36615e7a27e4cacdbc7d52fe07056db87e8b545f45b79dac1585288421cc40c8387a65afc5b0e7f2b95a68b3f106d1b76e9fcb5a42d339e031e77d0e767467b5aa2496ee8f3267cbb823168215852aa4ef';
```
ArnSch: Can this be a constant somewhere else and/or generated?

packages/dev-utils/src/constants.js
```diff
@@ -26,29 +32,24 @@ const constants = {
...
     K_MAX: 1048576,
-    /** Minimum value of an AZTEC Note
+    /** Maximum value that can be held in an AZTEC note during tests
      *  @constant K_MIN
      *  @type {string}
      *  @default 0
      */
     K_MAX_TEST: 14336,
-    /** Maximum value that can be held in an AZTEC note during tests
+    /** Minimum value that can be held in an AZTEC note
```
ArnSch, line 41: Introduce typo

packages/protocol/contracts/ACE/validators/joinSplitFluid/JoinSplitFluidABIEncoder.sol
```diff
@@ -156,7 +155,9 @@ library JoinSplitFluidABIEncoder {
             // 0x400 + metadataLength = byte length of output notes (0x120)
             // 0x420 + metadataLength = # of output notes (1)
             // 0x440 + metadataLength = offset to outputNotes[0] (0x60)
-            mstore(add(0x400, metadataLength), 0x120) // store length of output notes
+            let metadataLengthTemp := calldataload(add(metadataIndex, sub(metadata, 0x40)))
+
+            mstore(add(0x400, metadataLength), add(0x120, metadataLengthTemp)) // store length of output notes
```
ArnSch: All these adds are superfluous no?

ArnSch (changes requested):
WIP review, js interface should be simplified, and I believe you are doing quite a few `add`s which are all adding 0 to constants.

packages/aztec.js/test/encoder/inputCoder.js
thomas-waite: It already is - forgot to update this test with it. Will do

packages/dev-utils/src/constants.js
thomas-waite: Good spot, done

packages/aztec.js/src/encoder/outputCoder.js
thomas-waite: Good point, have done

packages/aztec.js/src/encoder/outputCoder.js
thomas-waite: Removed it and rolled the `decodeInputNote()` and `decodeOutputNote()` functions into one `decodeNote()`

packages/aztec.js/src/encoder/outputCoder.js
thomas-waite: The encoding functions are working from a slightly different input to the decoding functions, which means I don't believe the suggestions that work for the decoding functions will work for the encoding functions. 

In the decoding case, the function takes already correctly encoded notes i.e. `inputNotes` don't have any `metaData`, whilst `outputNotes` do have the `metaData`. So discriminating between input and output notes within the same decode function, based on the presence of `metaData` (and it's resulting impact on the size of the note) works. 

However in the encoding case, the function takes in unencoded notes still in the raw note schema format. At this stage, `inputNotes` and `outputNotes` are undifferentiated and, for example, both have `metaData`. So, discriminating between them based on the presence of `metaData` won't work, and there has to be a custom encode function for each type. It's the job of the encoder to format the raw notes correctly.

This encoding behaviour could be solved by defining a different note schema for `inputNotes` and `outputNotes`, but I think it's simpler to stick with one schema for all notes.

packages/aztec.js/src/proof/joinSplitFluid/index.js
thomas-waite: Yep, removed them

packages/protocol/contracts/ACE/validators/joinSplitFluid/JoinSplitFluidABIEncoder.sol
thomas-waite: yeah good catch, this was a short term debugging fix

ArnSch (approved):
LGTM!

PR #245, opened 2019-08-11
Feat add upgrade to ownable

ArnSch:
## Description

This PR does two things:

1. Adds a method callable by the owner of an Ownable zkasset which can be used to trigger the upgrade of the underlying note registry
2. Fixes two bugs which would have prevented further upgrades on note registries, and would have resulted in data loss. Test cases have been added for these bugs, making sure that the address stored in NoteRegistryManager for a particular note registry does not change on upgrade, and that the ownership of the proxy contract gets appropriately transferred on upgrade to the new factory.

The bugs in question could be found in the manager contract and the factory template. The bug causing any upgrade after the first to fail came from the fact that ownership of the proxy contract was not correctly transferred to the factory of the newly deployed behaviour. On subsequent upgrades, the factory contract notionally in charge of the proxy would not have the admin rights required to update the implementation the proxy used.

The second, and more serious bug, was caused by a re-assignment of the value stored in the registries mapping in the manager contract to be the newly deployed behaviour contract in an upgrade. On calls post-upgrade, the storage used in the transaction would be the storage of the behaviour contract instead of the proxy. 

## Testing instructions

`yarn test`

## Types of changes


thomas-waite (approved):
Looks good

PR #249, opened 2019-08-12
Fix burn note event log

thomas-waite:
## Description
Minor bug fix in `ZkAssetBurnable.sol` - thanks to @LeilaWang  for spotting!

Burn notes are treated as `inputNotes` and are removed from a note registry on successful verification of a burn proof. When this happens, events are emitted mark their destruction.

Each note should have a `DestroyNote()` event emitted. However, at the moment in the `confidentialBurn()` method, `logOutputNotes(burnedNotes)` is called - this results in `CreateNote()` events being emitted. 

This PR corrects this to `logInputNotes(burnedNotes)`, resulting in the correct emission of `DestroyNote()` events. 

## Checklist:


ArnSch (approved):
Makes sense - think I introduced this! Thanks for fixing and @LeilaWang for spotting.

PR #257, opened 2019-08-14
Feat move public permissioning to manager

ArnSch:
## Description

With upgradeability, we broke out the logic and storage of note registries into their own contracts, separate from ACE. To keep ACE as an abstraction layer and to make the client side code not have to keep track of where these extra contracts are, public tokens are still kept in ACE.

To deposit or withdraw, the previous logic required note registry contracts to call back to ACE with transfer instructions, but the permissioning logic on whether this transfer could happen was in the registry implementation contract. This opened up a fragility, where ACE had to assume and trust that any new upgrades of note registries were properly implemented, and didn't permit any malicious transaction to spend tokens which were not authorised to be spent.

This the level of fragility felt too high, so all permissioning logic and transfer logic has been repatriated in the NoteRegistryManager.sol contract, which is a part of ACE, and is non-upgradeable.

This also has a side-effect of simplifying both call flows and data structures in the manager contract and in the behaviour implementations.

## Testing instructions

`yarn test`

## Types of changes


thomas-waite (approved):
Looks good to me, good explanation in person

PR #261, opened 2019-08-15
fix(protocol): get rid of extranous vars+get rid of unlikely re-entrency

ArnSch:
## Description

Minor PR to get rid of a couple of unused variables, shadowed variables, and move some state update to the top of a fn after feedback from Slither.


thomas-waite (approved):
Looks good, as discussed in person

PR #301, opened 2019-08-29
Feat make zk asset support joinsplit upgrade

ArnSch:
## Description

This change fixes an issue with ZkAssets where, because of a hard coded proofId, any update to the JoinSplit validator could result in assets being locked out.

## Testing instructions

`yarn test`

## Types of changes

### DIFF
--- a/packages/protocol/contracts/ERC1724/base/ZkAssetMintableBase.sol
+++ b/packages/protocol/contracts/ERC1724/base/ZkAssetMintableBase.sol
@@ -61,6 +61,8 @@ contract ZkAssetMintableBase is ZkAssetOwnableBase {
     * @param _signatures - array of the ECDSA signatures over all inputNotes
     */
     function confidentialTransfer(uint24 _proofId, bytes memory _proofData, bytes memory _signatures) public {
+        bool result = supportsProof(_proofId);
+        require(result == true, "expected proof to be supported");
         // Check that it's a balanced proof
         (, uint8 category, ) = _proofId.getProofComponents();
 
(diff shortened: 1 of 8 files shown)


packages/protocol/contracts/ERC1724/base/ZkAssetMintableBase.sol
```diff
@@ -158,5 +147,35 @@ contract ZkAssetMintableBase is ZkAssetOwnableBase {
...
+    * @dev called when a mintable and convertible asset wants to perform an
+            action which puts the zero-knowledge and public
+            balance out of balance. For example, if minting in zero-knowledge, some
+            public tokens need to be added to the pool
+            managed by ACE, otherwise any private->public conversion runs the risk of not
+            having any public tokens to send.
+    *
+    * @param _value the value to be added
+    */
+    function supplementTokens(uint256 _value) internal {
```
thomas-waite, line 166: I like this, redefining the supplementTokens() related logic into the same function

packages/protocol/contracts/ERC1724/base/ZkAssetMintableBase.sol
```diff
@@ -49,21 +49,22 @@ contract ZkAssetMintableBase is ZkAssetOwnableBase {
...
+    * @param _proofData - bytes variable outputted from a proof verification contract, representing
+    * transfer instructions for the ACE
+    * @param _signatures - array of the ECDSA signatures over all inputNotes
     */
-    function confidentialTransfer(bytes memory _proofData, bytes memory _signatures) public {
+    function confidentialTransfer(uint24 _proofId, bytes memory _proofData, bytes memory _signatures) public {
+        // Check that it's a balanced proof
+        (, uint8 category, ) = _proofId.getProofComponents();
+
+        require(category == uint8(ProofCategory.BALANCED), "this is not a balanced proof");
```
thomas-waite, line 69: Good point, `confidentialTransfer()` in the `ZkAssetMintableBase.sol` should perform a check/some form of permissioning that the type of proof passed is supported by the asset

packages/protocol/contracts/ERC1724/base/ZkAssetMintableBase.sol
```diff
@@ -115,7 +120,7 @@ contract ZkAssetMintableBase is ZkAssetOwnableBase {
         bytes memory outputNotes,
         address publicOwner,
```
thomas-waite, line 126: Should `confidentialTransferFrom()` for the mintable ZkAsset also not have some form of permissioning around the type of proof it accepts? Should it check that the passed `uint24 _proof` is supported by the asset, like:
```
        bool result = supportsProof(_proofId);
        require(result == true, "expected proof to be supported")
```

We are overwriting and redefining `confidentialTransferFrom()` here, so the checks in the underlying `confidentialTransferFrom()` on the `ZkAssetOwnableBase.sol` are not being performed. 

We don't need to do this for the `ZkAssetBurnableBase.sol`  - it doesn't redefine the transfer methods

thomas-waite (changes requested):
Nice, only one significant comment - on the `ZkAssetMintableBase.sol`. I think we should add a check in that the proof type supplied to the `confidentialTransferFrom()` method is supported by the asset. 

Given that we are redefining and overwriting the `ZkAssetOwnableBase.sol` method, we no longer have the relevant check.

Other than that, looks good

packages/protocol/contracts/ERC1724/base/ZkAssetMintableBase.sol
ArnSch: Yep you're right!

ArnSch:
@thomas-waite ready to re-check

thomas-waite (approved):
Looks good

PR #307, opened 2019-09-02
Use real g2 point to test CRS

thomas-waite:
## Summary
This PR refactors the tests around checking that a fake trusted setup public key is rejected by the pairing check in our validator contracts. 

It takes a valid g2 point, one that is on the curve, and passes that in as the `FAKE_CRS`. Previously, just a random point was passed. This random point was likely not on the `bn128` curve and so the validators were throwing for a different reason, other than the pairing check rejecting a fake trusted setup.

## Checklist:


ArnSch (approved):
LGTM

PR #313, opened 2019-09-04
fix(aztec.js): adjust babel plugins to avoid unsafe-eval

LeilaWang:
## Description

1. Replace @babel/polyfill with @babel/plugin-transform-runtime
    We should not compile @babel/polyfill into the bundles as advised by [babel](https://babeljs.io/docs/en/babel-plugin-transform-runtime#why):

>  If you directly import core-js or @babel/polyfill and the built-ins it provides such as Promise, Set and Map, those will pollute the global scope. While this might be ok for an app or a command line tool, it becomes a problem if your code is a library which you intend to publish for others to use or if you can't exactly control the environment in which your code will run.

2. Add @babel/runtime to dependencies (Not devDependencies, since it is for the "runtime".)
    Doing so will prevent the async functions being transpiled into *regeneratorRuntime.wrap* (global scope).
    By using runtime transformer the code will be compiled to the following instead:
```
"use strict";
var _regenerator = require("@babel/runtime/regenerator");
var _regenerator2 = _interopRequireDefault(_regenerator);
function _interopRequireDefault(obj) {
  return obj && obj.__esModule ? obj : { default: obj };
}
var _marked = [foo].map(_regenerator2.default.mark);

function foo() {
  return _regenerator2.default.wrap(
    function foo$(_context) {
      // ...
    },
    _marked[0],
    this
  );
}
```

3. Don't init regeneratorRuntime
    Since we don't need the regeneratorRuntime variable anymore, we can remove its assignment in [regenerator](https://github.com/facebook/regenerator/blob/6e9e8d7747c2ab49927bdd9dd6261753181faec1/packages/regenerator-runtime/runtime.js#L714) by using webpack.DefinePlugin:
```
new webpack.DefinePlugin({
    'regeneratorRuntime': 'var unusedRegeneratorRuntime',
})
```

Which will change the code from

```
try {
  regeneratorRuntime = runtime; // undefined regeneratorRuntime
} catch (accidentalStrictMode) {
  Function("r", "regeneratorRuntime = r")(runtime);
}
```
to 
```
try {
  var unusedRegeneratorRuntime = runtime;
} catch (accidentalStrictMode) {
  Function("r", "regeneratorRuntime = r")(runtime);
}
```
The modified code will no longer trigger an error in strict mode. Where the Function call inside catch() will cause an _unsafe-eval_ error if forbid by CSP (eg: chrome/firefox extensions).

## Types of changes

Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/packages/aztec.js/babel.config.js
+++ b/packages/aztec.js/babel.config.js
@@ -1,0 +1,18 @@
+module.exports = {
+    // By default, babel assums all files are ES modules and use import to insert them.
+    // However, webpack doesn't allow import to be mixed with module.exports,
+    // which will cause this error:
+    //     Cannot assign to read only property 'exports' of object '#<Object>'
+    // So here we define sourceType = 'unambiguous' to tell babel to guess the file type.
+    //
+    // TODO - change sourceType to be 'module' (default value)
+    // This will require all files to use import/export
+    // and enforce babel to parse the files using the ECMAScript Module grammar.
+    // Under this mode, files are automatically strict and less likely to pollute global scope.
+    sourceType: 'unambiguous',
+    presets: ['@babel/preset-env'],
+    plugins: [
+        '@babel/plugin-transform-runtime',
+        '@babel/plugin-proposal-object-rest-spread',
+    ],
+};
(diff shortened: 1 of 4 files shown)


ArnSch (approved):
LGTM thank you for spotting

PR #315, opened 2019-09-05
Note viewing key permissioning system

thomas-waite:
## Summary
This PR implements a note viewing key permissioning system. 

It allows third parties to be granted access to a note, by the note owner adding an IES encrypted viewing key to the `metaData` of the note. 

A `noteAccess` mapping has been implemented to keep track of which Ethereum addresses have been granted view access to a note. Being granted approval gives that address permission to call `updateNoteMetaData()` - this allows the approved party to in turn grant view access to additional parties. 

The `noteAccess` mapping is based on the block timestamp. Specifically, it is a mapping of a `noteHash` to a mapping of `addressID` to `blockTimeStamp`:

```
    mapping(bytes32 => mapping(bytes32 => uint256)) public noteAccess;
``` 

`addressID` is `keccak256(approvedAddress, noteHash)`. 

In order to extract addresses from the `metaData` of a note, a `MetaDataUtils.sol` library has been written.

## Todo's
**1) Explicit permissioning step test**

The permissioning step in `ZkAssetBase.updateNoteMetaData()` needs explicitly testing. It's function is currently implied from querying the `noteAccess` mapping in certain scenarios - using a `ZkAssetTest` for which the permission requirement was removed. 

This was done due to the difficulty in creating a note - you need to know the public key - owned by a truffle account. Only truffle accounts can send transactions.

**2) Consider the implications of `updateNoteMetaDataI()` being called on `confidentialTotalMinted()` and `confidentialTotalBurned()` notes**

Consider the implications of these notes being passed and potential mitigations

## Testing instructions
Various test cases to check the permissioning flow have been added to `ZkAsset.js`.


packages/protocol/contracts/ERC1724/base/ZkAssetBase.sol
```diff
@@ -311,23 +317,73 @@ contract ZkAssetBase is IZkAsset, IAZTEC, LibEIP712 {
     * @param noteHash - hash of a note, used as a unique identifier for the note
     * @param metaData - metadata to update the note with
     */
-    function updateNoteMetaData(bytes32 noteHash, bytes calldata metaData) external {
+    function updateNoteMetaData(bytes32 noteHash, bytes memory metaData) public {
         // Get the note from this assets registry
         ( uint8 status, , , address noteOwner ) = ace.getNote(address(this), noteHash);
-        require(status == 1, "only unspent notes can be approved");
 
+        bytes32 addressID = keccak256(abi.encodePacked(msg.sender, noteHash));
```
ArnSch, line 320: This strikes me as superfluous if you're doing a two dimensional mapping.

Either do a mapping from `keccak256(abi.encodePacked(msg.sender, noteHash)) => uint256` or a two dimensional mapping from `noteHash => address => uint256`. Both will result in a unique timestamp per note per address.

packages/protocol/contracts/interfaces/IZkAsset.sol
```diff
@@ -15,7 +15,7 @@ contract IZkAsset {
         bool indexed _canAdjustSupply,
         bool _canConvert
     );
-    event ApprovedAddress(address indexed addressApproved, bytes32 indexed noteHash);
+    event ApprovedAddress(address indexed addressApproved, bytes32 indexed noteHash, address contractAddress);
```
ArnSch: Should contractAddress be indexed to make finding all approved notes for a user for an asset easier?

packages/protocol/contracts/libs/MetaDataUtils.sol
```diff
@@ -8,23 +8,35 @@ pragma solidity >=0.5.0 <= 0.6.0;
...
+    */
+    function extractAddress(bytes memory metaData, uint256 addressPos) public returns (address desiredAddress) {
+        /**
+        * Memory map of metaData
+        * 0x00 - 0x20 : length of metaData
+        * 0x20 - 0x81 : ephemeral key
+        * 0x81 - 0xa1 : approved addresses offset
+        * 0xa1 - 0xc1 : encrypted view keys offset
+        * 0xc1 - 0xe1 : app data offset
+        * 0xe1 - L_addresses : approvedAddresses
```
ArnSch, line 28: Where is L_addresses set? I assume that it's the first word at 0xe1, but would be good to document.

packages/protocol/contracts/libs/MetaDataUtils.sol
```diff
@@ -8,23 +8,35 @@ pragma solidity >=0.5.0 <= 0.6.0;
...
+        * (0xe1 + L_addresses) - (0xe1 + L_addresses + L_encryptedViewKeys) : encrypted view keys
+        * (0xe1 + L_addresses + L_encryptedViewKeys) - (0xe1 + L_addresses + L_encryptedViewKeys + L_appData) : appData
```
ArnSch, line 30: ditto for L_encryptedViewKeys

packages/protocol/contracts/libs/MetaDataUtils.sol
```diff
@@ -8,23 +8,35 @@ pragma solidity >=0.5.0 <= 0.6.0;
...
+        */
         assembly {
-            addressToApprove := mload(add(metaData, 0x20))
+            desiredAddress := mload(
+                add(
+                    add(
+                        metaData,
+                        add(0xe1, 0x20)  // go to the start of addresses, jump over first word
```
ArnSch, line 40: I think it's worth the overhead to sense check that you're not going out of the bounds of `0xe1 + L_addresses`. Could be as simple as checking that addressPos < L_addresses.

packages/protocol/test/ERC1724/ZkAsset.js
```diff
@@ -330,7 +334,7 @@ contract('ZkAsset', (accounts) => {
             expect(receipt.status).to.equal(true);
         });
 
-        it('should update the metadata of a note when ZkAsset.updateNoteMetadData() called', async () => {
+        it('should update the metadata of a note when ZkAsset.updateNoteMetaData() called', async () => {
             /**
              * Using ZkAssetTest as the ZkAsset here. ZkAssetTest has the safety feature
```
ArnSch: I don't feel comfortable with this solution. You can get the accounts used in ganache from a set mnemonic (see @charlie-g-cowan's branch).

This isn't really testing the zkAsset implementation of updateNoteMetadata. As it stands, we could drastically change that method implementation in zkAssetBase and this test would still pass.

packages/protocol/test/ERC1724/ZkAsset.js
```diff
@@ -343,43 +347,57 @@ contract('ZkAsset', (accounts) => {
...
+            await zkAsset.methods['confidentialTransfer(bytes,bytes)'](data, signatures, { from: accounts[0] });
+
+            const { noteHash } = depositOutputNotes[0];
+            const metaDataTimeLog = await zkAsset.metaDataTimeLog.call(noteHash);
+            expect(metaDataTimeLog).to.not.equal(undefined);
+        });
+
+        it('should approve and grant addresses noteAccess', async () => {
             // create one outputNote, then place other addresses on the approved noteAccess mapping
             const zkAssetTest = await ZkAssetTest.new(ace.address, erc20.address, scalingFactor);
```
ArnSch, line 450: Ditto on zkAssetTest

packages/protocol/contracts/test/ERC1724/ZkAssetTest.sol
```diff
@@ -20,22 +20,28 @@ contract ZkAssetTest is ZkAsset {
...
     * @param noteHash - hash of a note, used as a unique identifier for the note
     * @param metaData - metadata to update the note with
     */
-    function updateNoteMetaData(bytes32 noteHash, bytes calldata metaData) external {
-        // Get the note from this assets registry
+    function updateNoteMetaData(bytes32 noteHash, bytes memory metaData) public {
         ( uint8 status, , , address noteOwner ) = ace.getNote(address(this), noteHash);
-        require(status == 1, "only unspent notes can be approved");
+        (, , bytes32 confidentialTotalMinted,
+        bytes32 confidentialTotalBurned, , , , bool canAdjustSupply) = ace.getRegistry(address(this));
```
ArnSch:
```suggestion
        (
            ,,
            bytes32 confidentialTotalMinted,
            bytes32 confidentialTotalBurned
            ,,,,
            bool canAdjustSupply
        ) = ace.getRegistry(address(this));
```

ArnSch (changes requested):

packages/protocol/contracts/interfaces/IZkAsset.sol
thomas-waite: Chatted with the extension team about this to decide which fields should be indexed and arrived at the above. Will double check with them and mention your point, makes sense

thomas-waite: Just spoke to Joe. They didn't need `contractAddress` indexing and now also don't need this event - so removing the event altogether

packages/protocol/contracts/libs/MetaDataUtils.sol
thomas-waite: `L_addresses` isn't so much as set, rather it's just the length in memory that the addresses take up. The first word at `0xe1` is the number of addresses.

The extension has a variety of `metaData` encoding functions, which take a user's `metaData` intended for a note, and encodes it according to this ABI outlined in `MetaDataUtils.sol` . Will add a comment explaining this

thomas-waite: The first word at `0xe1` is the number of addresses. `L_addresses` isn't so much as set, rather it's just the length in memory that the addresses take up

packages/protocol/contracts/libs/MetaDataUtils.sol
thomas-waite: Same as above

packages/protocol/contracts/ERC1724/base/ZkAssetBase.sol
thomas-waite: Good point, my bad. It should just be a mapping of `keccak256(abi.encodePacked(msg.sender, noteHash)) => uint256`

packages/protocol/test/ERC1724/ZkAsset.js
thomas-waite: Fully agree, the current testing setup is not good enough. This is one of the todo's I mentioned for the WIP PR - was implementing after this review

packages/protocol/test/ERC1724/ZkAsset.js
thomas-waite: Same as above

packages/protocol/contracts/libs/MetaDataUtils.sol
ArnSch: Yes, I more mean documenting that the first word at 0xe1 is in fact num of addresses, same for L_encryptedViewKeys

thomas-waite: Gotcha - this is standard across our ABI encoding. The first word of any dynamic bytes array, within another dynamic bytes array, is the number of elements in that array. We should document and make this ABI encoding clear

packages/protocol/contracts/test/ERC1724/ZkAssetTest.sol
thomas-waite: Done

packages/protocol/contracts/libs/MetaDataUtils.sol
thomas-waite: Sure, good idea - done

ArnSch (approved):
LGTM

PR #324, opened 2019-09-13
Re-add pairing check in swap validator

thomas-waite:
## Summary
This PR re-adds the pairing check back into the `Swap.sol` validator contract, necessary to ensure commitments are valid.

## Checklist:

### DIFF
--- a/packages/aztec.js/src/proof/swap/verifier.js
+++ b/packages/aztec.js/src/proof/swap/verifier.js
@@ -24,20 +24,34 @@ class SwapVerifier extends Verifier {
         challengeResponse.appendBN(this.proof.sender.slice(2));
         challengeResponse.data = [...challengeResponse.data, ...rollingHash.data];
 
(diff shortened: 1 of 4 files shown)


packages/aztec.js/src/proof/swap/verifier.js
```diff
@@ -24,20 +24,34 @@ class SwapVerifier extends Verifier {
...
+        let reducer = bn128.zeroBnRed;
         this.data.forEach((note, i) => {
+            let aBar;
             let kBar;
-            const { aBar, gamma, sigma } = note;
+            let c;
+            const { gamma, sigma } = note;
 
-            if (i <= 1) {
+            if (i === 0) {
```
ArnSch, line 34: Raising that we could check only if i == 1 and i == 2, since we only care about the pairing on output notes.

ArnSch (approved):
Apart from optimisation point, LGTM. Merging to add this work into recursive hashing fix.

PR #326, opened 2019-09-17
Feat proof versioning

thomas-waite:
## Summary
Implement a proof construction versioning system in the `aztec.js` package. 

## Description
The motivation of this PR was to develop a systematic method whereby additional proofs could be added to `aztec.js`, without making any breaking changes to the API. 

### Individual proof export scripts
Each proof now has it's own proof export script. The main method in each script is a function with the same name as the proof, to prevent any breaking changes. The API to construct a proof is the same as before:
```
const { JoinSplitProof } = require('aztec.js');

const proof = new JoinSplitProof(...args);
```

There is however now also the option to specify a version of a proof from a particular epoch, such as:

```
const proof = new JoinSplitProof.epoch(2)(...args);
```

### proofHandler.js
This is a generic module that is used to manage the process of exporting individual proofs, and is relied on for the 'backend' of each individual proof export script. 

It also enables additional functionality that allows a user to specify a `defaultEpochNum` which a user wants to use proofs from. 

### proofCatalog.js
Closely related to the `proofHandler` and required for it's functionality is the `proofCatalog.js` script. This contains a mapping between epoch numbers and the proof versions, which is queried by the `proofHandler` when the user calls a proof.

Using such a mapping to keep track of which proofs belong to which epoch numbers will make this versioning system more scalable in the future - additional proofs and epochs can easily be added.  

## Testing instructions
It is important that the `proofHandler` behaves as expected. As such various success and failure tests have been added to confirm behaviour.


packages/aztec.js/src/proof/proofCatalog.js
```diff
@@ -0,0 +1,48 @@
...
+const PrivateRangeProof66562 = require('./epoch0/UTILITY/privateRange');
+const PublicRangeProof66563 = require('./epoch0/UTILITY/publicRange');
+const SwapProof65794 = require('./epoch0/BALANCED/swap');
+
+const proofCatalog = {};
+
+/*
+ * Default epoch number for which proof construction methods are exported
+ */
+proofCatalog.defaultEpochNum = 1;
```
ArnSch: The way this is currently structured, it assumes that we will never add an epoch on one proof without also adding proofs for all the existing proof types for that new epoch. Is that true?

packages/aztec.js/test/proof/dividend/verifier.js
```diff
@@ -7,7 +7,7 @@ const { padLeft, randomHex } = require('web3-utils');
 const sinon = require('sinon');
 
 const { DividendProof } = require('../../../src/proof');
-const DividendVerifier = require('../../../src/proof/dividend/verifier');
+const DividendVerifier = require('../../../src/proof/epoch0/UTILITY/dividend/verifier');
```
ArnSch: small, but I think it makes sense for the proof type to be a parent of an epoch:

```
- UTILITY
  - epoch0
  - epoch1
```
instead of 
```
- Epoch0
  - utility
  - balanced
- Epoch1
  - utility
```

packages/aztec.js/test/proof/joinSplitFluid/verifier.js
```diff
@@ -1,112 +1,112 @@
...
-                expect(verifier.errors.length).to.equal(3);
-                expect(verifier.errors[0]).to.equal(errors.codes.SCALAR_IS_ZERO);
-                expect(verifier.errors[1]).to.equal(errors.codes.INCORRECT_NOTE_NUMBER);
-                expect(verifier.errors[2]).to.equal(errors.codes.CHALLENGE_RESPONSE_FAIL);
-                validateInputs.restore();
-            });
-        });
-    });
-});
+// const { errors } = require('@aztec/dev-utils');
```
ArnSch: Do you mean for this to be commented?

ArnSch (changes requested):
Small comments around relative hierarchy of proof types vs epochs, but otherwise 👍

packages/aztec.js/test/proof/dividend/verifier.js
thomas-waite: As discussed, this makes sense

packages/aztec.js/test/proof/joinSplitFluid/verifier.js
thomas-waite: Good point, not sure how I missed that - should be uncommented. Fixed

packages/aztec.js/src/proof/proofCatalog.js
thomas-waite: As discussed earlier, good point. We're going to add another mapping whereby each proof type has a default proof epoch, which the user can set

thomas-waite:
Ready for a re-check

ArnSch (approved):

PR #331, opened 2019-09-19
Fix confidential approve

thomas-waite:
## Summary
Fix `confidentialApprove()` so that signature replay attacks can not be used to re-grant a spender note spending access once their permission has been revoked. 

## Description
The main fix is that a mapping `signatureLog` has been added to the `ZkAssetBase.sol` along with corresponding validation logic in `confidentialApprove()`. 

The mapping maps a signature hash to a bool, with the bool representing whether the signature has previously been presented:

```
mapping(bytes32 => bool) signatureLog
```

Whenever a signature is presented to `confidentialApprove()`, the following check is first performed to see if the hash of the signature is present in `signatureLog`. If it is, the transaction reverts.

If not, the transaction proceeds and an hash(signature):true entry is made to `signatureLog`.  

The `signNoteForConfidentialApprove()` method now also takes a `spenderApproval` variable rather than note`status`. In `confidentialApprove` there is a check to make sure that `status` is 1, so adding to the signature is redundant. Adding the `spenderApproval` variable ensures that a particular signature can only be used to either 1) grant permission, or 2) revoke permissionl

### Efficiency considerations
Granting approval using a `confidentialApprove()` is likely to be one of the most common `ZkAsset` functions called, therefore gas efficiency is key. 

I believe using a mapping for the permissioning is one of the most gas efficient ways to do this. Alternatives, such as storing presented signatures in an array and iterating through each time to check if the signature was already presented, are far more expensive. 

## Testing instructions
Additional tests have been added to `ZkAssetOwnable.js`:

_Success tests_
- 'should revoke confidentialApprove() permission granting': confirm that the revoke permission behaviour of `confidentialApprove()` works as expected

_Failure tests_
- 'should fail to perform a replay attack, by using a previous signature in confidentialApprove()': check that the fix prevents a replay attack

### DIFF
--- a/packages/aztec.js/src/signer/index.js
+++ b/packages/aztec.js/src/signer/index.js
@@ -53,17 +53,18 @@ signer.generateZKAssetDomainParams = (verifyingContract) => {
  * @param {string} verifyingContract address of target contract
  * @param {string} noteHash noteHash of the note being signed
  * @param {string} spender address of the note spender
+ * @param {bool} spenderApproval boolean determining whether the spender is being granted approval
+ * or revoked
  * @param {string} privateKey the private key of message signer
  * @returns {string} ECDSA signature parameters [r, s, v], formatted as 32-byte wide hex-strings
  */
-signer.signNoteForConfidentialApprove = (verifyingContract, noteHash, spender, privateKey) => {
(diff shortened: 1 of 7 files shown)


ArnSch (approved):
LGTM

PR #334, opened 2019-09-25
Fix recursive hashing

ArnSch:
## Summary

Move away from recursive hashing in the pairing optimisation in favour of exponentiating a hash to avoid an unlikely pre-image attack.

## Testing instructions

`yarn test`

## Types of changes


packages/aztec.js/src/proof/dividend/index.js
```diff
@@ -59,11 +59,10 @@ class DividendProof extends Proof {
                 // bk_3 = (z_b)(bk_1) - (z_a)(bk_2)
                 bk = zbRed.redMul(blindingScalars[0].bk).redSub(zaRed.redMul(blindingScalars[1].bk));
             }
-
-            const xbk = bk.redMul(reducer); // xbk = bk*x
-            const xba = ba.redMul(reducer); // xba = ba*x
+            const x = reducer.redPow(new BN(i + 1));
```
thomas-waite: `x` and `reducer` are two words for the same thing. Small thing, but it might give the impression that they're representing actually different entities. 

Why not do something like:
```
let initialReducer = this.rollingHash.redKeccak(); // "x" in the white paper 
```
and then redefine as:

```
reducer = initialReducer.redPow(new BN(i + 1)); 
```

packages/aztec.js/src/proof/dividend/verifier.js
```diff
@@ -36,24 +36,26 @@ class DividendVerifier extends Verifier {
 
         challengeResponse.data.push(...rollingHash.data);
 
-        let reducer = rollingHash.redKeccak(); // "x" in the white paper
+        const reducer = rollingHash.redKeccak(); // "x" in the white paper
         this.data.forEach((item, i) => {
+            const x = reducer.redPow(new BN(i + 1));
```
thomas-waite: same question about using two names, `reducer` and `x`, for the same variable

thomas-waite (approved):
Looks good to me. 

Have one small comment in the proof construction scripts, around using two variable names for the same entity

thomas-waite (approved):
Looks good!

PR #340, opened 2019-09-30
fix(protocol): add validation on multisig

ArnSch:
## Summary

Make sure that provided destination is a contract.


thomas-waite (approved):
Looks good

PR #344, opened 2019-10-02
batchConfidentialApprove() on ZkAsset

thomas-waite:
## Summary
This PR adds a `batchConfidentialApprove()` method to the `ZkAssetBase.sol` contract, to allow multiple notes to be approved to be spent by a third party address in one function call. 

This improves the UX for users.

### aztec.js
A new EIP712 signing function has been added to the `signer` module:
```
signNotesForBatchConfidentialApprove(verifyingContract, noteHashes, spender, spenderApprovals, privateKey)
```

`noteHashes` is an array of the hashes of all notes being simultaneously approved or revoked. `spenderApprovals` is an array of booleans representing whether each note in the corresponding index position in the `noteHashes` array is having permission approved or revoked. 

### protocol
The `ZkAssetBase.sol` contract has a new method `batchConfidentialApprove()`. This validates the signature generated by the new signing function, and will update the mapping `confidentialApproved` with the resulting permissioning. 

This function includes signature replay protection, checks that the notes in question are valid and that the signature is valid - i.e. that the owner of the notes in question is responsible for the supplied signature. 

### dev-utils
The associated EIP712 signature schemas and hashes have been added.

## Testing instructions
Various tests confirming the `aztec.js` signing function and contract signature validation have been added. 

## Types of changes
- New feature: `batchConfidentialApprove()`

### DIFF
--- a/packages/aztec.js/src/signer/index.js
+++ b/packages/aztec.js/src/signer/index.js
@@ -73,6 +73,36 @@ signer.signNoteForConfidentialApprove = (verifyingContract, noteHash, spender, s
 };
 
 /**
+ * Create an EIP712 ECDSA signature over an array of AZTEC notes, used to either grant or revoke delegation of note control
+ * to a third party using the batchConfidentialApprove() of a ZkAsset.
+ *
+ * ECSA signature is formatted as follows:
+ * - `r` and `s` occupy 32 bytes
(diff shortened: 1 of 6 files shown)


ArnSch (approved):
👍

PR #358, opened 2019-10-14
Make ZERO_VALUE_NOTE_HASH updateable

thomas-waite:
## Summary
Make `ZERO_VALUE_NOTE_HASH` updateable, so that `mint` and `burn` functionality works if `ACE` is updated with a new/updated `CRS`.

## Description
`ZERO_VALUE_NOTE_HASH` is a constant which represents the hash of a note with `k = 0` and `a = 1`. It's purpose is to serve as the initial total minted/burned note when an adjustable note registry is created. 

If the `CRS` is updated in `ACE` by calling `ace.setCommonReferenceString()`, then this would result in `mint` and `burn` functionality being disabled. The `ZERO_VALUE_NOTE_HASH` would now not match that used in `aztec.js`  proof construction, as the two would have been created over different sets of elliptic curve data points. 

The `ZERO_VALUE_NOTE_HASH` has been removed from the `IAZTEC.sol` contract to decouple it from `ACE`, which is non-upgradeable. 

Instead, it has been moved into the contract where it is used - the `Behaviour.sol` contracts of the note registries. This means that in the event of a `CRS` update, note registries can be made compatible and have access to `mint()` and `burn()` by deploying an upgraded `Behaviour` contract.

## Testing instructions
A success test confirming that the `ZERO_VALUE_NOTE_HASH` can be successfully updated has been added to the test script `ACE.js`.

## Types of changes

### DIFF
--- a/packages/protocol/contracts/ACE/ACE.sol
+++ b/packages/protocol/contracts/ACE/ACE.sol
@@ -34,6 +34,7 @@ contract ACE is IAZTEC, Ownable, NoteRegistryManager {
         address validatorAddress
     );
     event IncrementLatestEpoch(uint8 newLatestEpoch);
+    event SetZeroValueNoteHash(bytes32 zeroValueNoteHash);
 
     // The commonReferenceString contains one G1 group element and one G2 group element,
     // that are created via the AZTEC protocol's trusted setup. All zero-knowledge proofs supported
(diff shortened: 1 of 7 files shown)


ArnSch (approved):
Good change. As discussed, we should move CRS out of ACE too.

PR #369, opened 2019-11-06
Replace CRS with original hosted on cloudfront

thomas-waite:
## Summary
This PR replaces the CRS used by the repo with the original one that is hosted by CloudFront. This is necessary so that when notes are created in the SDK/Extension (which ping CloudFront) rather than use the `localSetupDatabase`, then they are able to be used with the `protocol`. 

Replacing the CRS in the repo involved replacing the `localSetupDatabase` with a selection of files from CloudFront, updating `t2` in the `configCRS.js` scripts, as well as the `h` point.


ArnSch (approved):
LGTM!

PR #380, opened 2019-11-15
Integration testing harness

thomas-waite:
## Summary
This PR adds a testing harness to enable simpler, faster and cheaper integration tests on testNets/mainNet. 

It introduces a new private package, `integration-test`, which uses a `setup` class to create the necessary environment for the integration test. The tests make use of the contracts already deployed and recorded in the `contract-artifacts` and `contract-addresses` packages, in order to avoid the significant costs of having to redeploy on each test.  

This work has been done with the view that our integration tests will involve in the future, with new functionality and tests being added.

## Description
A private `integration-test` package has been created. This runs an integration test, called `keyFlows.js` which tests critical AZTEC functionality such as:
- proof verification
- minting and burning
- confidentialTransfer()
- confidentialTransferFrom()
- confidentialApprove()
- upgradeability

It takes a `Setup` class instance to create the relevant environment for the test.

## Testing instructions
`cd` into the `integration-test` directory, and then do one of the following depending on desired functionality:

1) `yarn test:testNets` - run both ropsten and rinkeby tests
2) `yarn test:ropsten` - run tests on ropsten
3) `yarn test:rinkeby` - run tests on rinkeby

## Types of changes

New feature (non-breaking change which adds functionality)
- Testing harness for integration tests on testNets/mainNet

## Future direction/next steps
This testing harness could be improved by adding the following functionality:
- **Automated testNet ether faucet minting:** there are now two testing accounts that need to be maintained

- **Automated tenderly integration:** the contract artifacts should be automatically pushed using `tenderly push` on build, rather than having to manually do it. The `contract-artifacts` need saving with the deployed network information to enable automation

- **Add support for minting/burning tests beyond initial mint/burn** - in order to mint/burn repeatedbly, the total minted/burned needs to be kept track of and the representative note rederived for use in mocha tests

- **Possible internal node setup**


.circleci/config.yml
```diff
@@ -132,6 +143,14 @@ workflows:
                 filters:
                     branches:
                         ignore: gh-pages
+            - testNet:
+                requires:
+                    - build
+                filters:
+                    branches:
+                        ignore: gh-pages
+                #     - lint
```
ArnSch: Meant to leave this as comments?

packages/integration-test/scripts/test.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+echo "start of test script"
+# Exit script as soon as a command fails.
+set -o errexit
+echo "after second command"
```
ArnSch: Cleanup

packages/integration-test/src/config.js
```diff
@@ -0,0 +1,68 @@
...
+    'JoinSplitFluid',
+    'PrivateRange',
+    'PublicRange',
+    'Swap',
+    'ZkAssetAdjustable',
+];
+
+/**
+ * @constant {string} NETWORK - name of the test network for which the integration test is to be run.
+ * Extracted from the flag passed to the truffle test command - it is the last flag to be passed
```
ArnSch:
```suggestion
 * Extracted from the arguments passed to the truffle test command - it is the last argument to be passed
```
This is really pedantic, but flags in unix != arguments

packages/integration-test/src/setup.js
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * Sets up the environment for an integration test. Specifically, it creates Truffle Contract instances
+     * from the aztec contract-artifacts; at the appropriate address in the contract-addresses package.
+     *
+     * Can be used to setup a local development testing environment for an integration test
+     *
+     * @param {Object} config - config used to modify the test environment setup
+     * @param {String} accounts - Ethereum addresses to be used in the test
+     */
+    constructor(config, accounts) {
```
ArnSch: The config situation can be simplified, by having the constructor take in an optional object with the config values, and otherwise having defaults as state in the class. Would fit the idea of having data + related logic in one place.

packages/integration-test/src/setup.js
```diff
@@ -0,0 +1,174 @@
...
+        this.contractsToDeploy = config.contractsToDeploy;
+
+        this.provider = web3.currentProvider;
+        this.accounts = accounts;
+
+        const networks = ['ropsten', 'rinkeby', 'kovan', 'mainnet'];
+        if (networks.includes(this.NETWORK)) {
+            console.log('Configuring integration test for', this.NETWORK, 'network');
+            this.networkId = networkIDs[capitaliseFirstChar(this.NETWORK)];
+            this.getAddresses();
```
ArnSch: Minor, but from a clarity point of view, would be nice if assignment to class variables was done in constructor.
```suggestion
            this.contractAddresses = this.getAddresses();
```
with this.getAddresses() simply returning data instead of doing assignment. Would make these initialisation fns closer to getters, with no or few unexpected side-effects.

ArnSch: This would also mean you could expose a getter to the end user of the class, and it would perform as expected (return the data) instead of modifying the instance you're working with.

packages/integration-test/src/setup.js
```diff
@@ -0,0 +1,174 @@
...
+     * returns the Truffle representation of the contract
+     * @param {string} nameOfContract
+     */
+    getContractPromise(nameOfContract) {
+        const contractObject = {};
+        let extractedContractArtifact;
+        let extractedContractAddress;
+
+        try {
+            extractedContractArtifact = contractArtifacts[nameOfContract];
```
ArnSch: Does this ever throw? Unless I'm missing something, this should just return `undefined` in the event that `nameOfContract` is not in object?

packages/integration-test/src/setup.js
```diff
@@ -0,0 +1,174 @@
...
+        } catch (err) {
+            throw new AztecError(codes.NO_CONTRACT_ARTIFACT, {
+                message: 'Requested contract does not exist in @aztec/contract-artifacts',
+                requestedContract: nameOfContract,
+            });
+        }
+
+        try {
+            extractedContractAddress = this.contractAddresses[nameOfContract];
```
ArnSch: Ditto

packages/integration-test/src/setup.js
```diff
@@ -0,0 +1,174 @@
...
+            throw new AztecError(codes.NO_CONTRACT_ADDRESS, {
+                message: 'Requested contract does not have a deployed address in @aztec/contract-addresses',
+                requestedContract: nameOfContract,
+            });
+        }
+
+        contractObject[nameOfContract] = async () => {
+            const truffleRepresentation = this.getTruffleContractRepresentation(extractedContractArtifact);
+            return truffleRepresentation.at(extractedContractAddress);
+        };
```
ArnSch: I think it might be cleaner to have this function simply return that `async () => {}`, and construct the object you're constructing in caller.

(i.e. in `getContractPromises()`, do something like: 
```
this.contractPromises = this.contractsToDeploy.map((nameOfContract) => ({ [nameOfContract]: this.getContractPromise(nameOfContract)) });
```

)

packages/integration-test/src/setup.js
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /**
+     * @method getContracts - get the JavaScript objects representing contracts deployed at a specific address, for use
+     * in testing
+     * @returns {Object} Truffle contracts, representing the contract deployed at a specific address
+     */
+    async getContracts() {
+        const allContractObjects = (await Promise.all(
+            this.contractPromises.map(async (currentContractPromise, index) =>
```
ArnSch: At this point `this.contractPromises` is an array of objects, each object having a contract name as a key, and a promise as a value. You currently don't make use of this data.

You can restructure `this.contractPromises` as an array of arrays, each inner array being `[contractName, promise]`, which would make both your map and your reduce simpler.

packages/integration-test/src/utils.js
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Make the first letter of a given single word input string capitalised
+ *
+ * @capitaliseFirstChar
+ * @param {string} stringToCapitalise - input for which the first letter should be capitalised
+ *
+ * @returns {string} - input string with first letter capitalised
+ */
+utils.capitaliseFirstChar = (stringToCapitalise) => {
+    return stringToCapitalise.charAt(0).toUpperCase() + stringToCapitalise.slice(1);
```
ArnSch: More informational: Strings are arrays of chars. You can access a char by just using square brackets.

packages/integration-test/src/utils.js
```diff
@@ -0,0 +1,43 @@
...
+ * Generate a factory ID based on three supplied uint8's: epoch, cryptoSystem and
+ * assetType
+ *
+ * @method generateFactoryId
+ * @param {Number} epoch - uint8 representing factory version control
+ * @param {Number} cryptoSystem - uint8 representing the cryptosystem the factory is associated with
+ * @param {Number} assetType - uint8 representing the type of the asset i.e. is it convertible,
+ * adjustable
+ */
+utils.generateFactoryId = (epoch, cryptoSystem, assetType) => {
```
ArnSch, line 39: Is this duplicated from somewhere or is this the de-duplication? Sorry if latter!

packages/integration-test/test/integration.js
```diff
@@ -0,0 +1,393 @@
...
+            FactoryAdjustable201907: upgradeFactory,
+            JoinSplit: joinSplitValidator,
+            JoinSplitFluid: joinSplitFluidValidator,
+            PrivateRange: privateRangeValidator,
+            PublicRange: publicRangeValidator,
+            Swap: swapValidator,
+            ZkAssetAdjustable: zkAsset,
+        } = await setup.getContracts());
+
+        ({ sender, publicOwner, delegatedAddress, opts } = setup.getTransactionTestingAddresses());
```
ArnSch, line 62: This is a good example of comment about making getters not do assignment: you've modified the instance your working with, is that expected?

packages/integration-test/truffle-config.js
```diff
@@ -0,0 +1,113 @@
...
+        console.log('Please set a MNEMONIC in a .env file');
+        process.exit(1);
+    }
+    if (!process.env.INFURA_API_KEY) {
+        console.log('Please set your INFURA_API_KEY');
+        process.exit(1);
+    }
+    return () => {
+        return new HDWalletProvider(
+            process.env.MNEMONIC,
```
ArnSch, line 30: This is probably why CircleCI is failing.

packages/integration-test/truffle-config.js
```diff
@@ -0,0 +1,113 @@
...
+        development: {
+            provider: engine,
+            gas: 6500000,
+            gasPrice: toHex(toWei('1', 'gwei')),
+            network_id: '*', // eslint-disable-line camelcase
+            port: 8545,
+        },
+        mainnet: {
+            provider: mainnetProvider,
+            gas: 6000000,
```
ArnSch: These aren't current anymore, we should update

ArnSch (changes requested):
I think the Setup class can be cleaned up, otherwise great to have this done!

packages/integration-test/src/setup.js
ArnSch: The suggestion will change a bit if you implement my next comment :)

.circleci/config.yml
thomas-waite: part of my WIP circleci debugging - will clean up before merged

packages/integration-test/scripts/test.sh
thomas-waite: same as above

packages/integration-test/src/config.js
thomas-waite: Good point, done

packages/integration-test/src/setup.js
thomas-waite: Yep, very good point. Implemented

packages/integration-test/src/utils.js
thomas-waite: Didn't realise, nice

packages/integration-test/test/integration.js
thomas-waite: You're right, and you wouldn't expect this function to do that. Moved modifying the state to the constructor

packages/integration-test/truffle-config.js
thomas-waite: Yeah, think so. Should work now we've updated

packages/integration-test/truffle-config.js
thomas-waite: Good point, have updated with latest figures

packages/integration-test/src/setup.js
thomas-waite: You're right. Have refacted to check if `undefined` and throw if so

packages/integration-test/src/setup.js
thomas-waite: Same fix as above

packages/integration-test/src/setup.js
thomas-waite: Yep, this is a really good idea. Have implemented

packages/integration-test/src/setup.js
thomas-waite: Done

packages/integration-test/src/utils.js
thomas-waite: Duplicated, from test helpers in `protocol`. Not straightforward to resolve, without moving the various helpers into `dev-utils` or something similar. Leaving as is for time being

packages/integration-test/src/setup.js
```diff
@@ -0,0 +1,242 @@
...
+
+        if (networks.includes(this.NETWORK)) {
+            console.log('configuring integration test for', this.NETWORK, 'network');
+
+            const [sender, delegatedAddress] = accounts;
+            const publicOwner = sender;
+            const opts = { from: sender };
+
+            this.accounts = accounts;
+            this.contractsToDeploy = this.config.contractsToDeploy;
```
ArnSch: Very minor, but you can probably skip these `.config.` reassignments no?

packages/integration-test/src/setup.js
```diff
@@ -0,0 +1,242 @@
...
+
+    /**
+     * @method getContracts - generate and get the JavaScript objects representing contracts deployed at a specific address,
+     * for use in testing
+     * @returns {Object} Truffle contracts, representing the contract deployed at a specific address. The keys are the contract names,
+     * the values are the Truffle contract at a particular address
+     */
+    async getContracts() {
+        const allContractObjects = (await Promise.all(
+            this.contractPromises.map(async (currentContractPromise) => Promise.all(currentContractPromise)),
```
ArnSch: Why the `Promise.all` inside the fn?

packages/integration-test/src/setup.js
```diff
@@ -0,0 +1,242 @@
...
+        const extractedContractAddress = this.contractAddresses[nameOfContract];
+        if (extractedContractAddress === undefined) {
+            throw new AztecError(codes.NO_CONTRACT_ADDRESS, {
+                message: 'Requested contract does not have a deployed address in @aztec/contract-addresses',
+                requestedContract: nameOfContract,
+            });
+        }
+
+        const truffleRepresentation = this.getTruffleContractRepresentation(extractedContractArtifact);
+        return truffleRepresentation.at(extractedContractAddress); // this is going to return a promise
```
ArnSch: Would be cleaner to return an `async` function which can be evaluated and awaited. Right now, unless am missing something with this code, you will start evaluating `truffleRepresentation.at(...)` as soon as this fn hits this line. You happen to be saving the `Promise` which this evaluation returns while waiting for the async fetch to actually return data, but this line already adds instructions to the event loop.

All in all, not catastrophic given this is only a fetch. If this call had side effects, you could for example unwittingly be writing multiple times.

packages/integration-test/src/setup.js
```diff
@@ -0,0 +1,242 @@
...
+             * @param {BN} scalingFactor - factor to convert between AZTEC note value and ERC20 token value
+             */
+            scalingFactor: ERC20_SCALING_FACTOR,
+        };
+
+        if (config) {
+            this.config = config;
+        } else {
+            this.config = defaultConfig;
+        }
```
ArnSch: Would replace the above with: 
```suggestion
        this.config = Object.assign({}, defaultConfig, config)
```
Otherwise, someone passing in a config objet needs to pass in every single config key. The above will use the default config, only overwriting the keys passed in through `config`

ArnSch (changes requested):
Looks great, save for one anti-pattern. Happy to talk through it if comment is confusing.

packages/integration-test/src/setup.js
thomas-waite: yep good point, have done

packages/integration-test/src/setup.js
thomas-waite: That's a really good idea, have done

packages/integration-test/src/setup.js
thomas-waite: Very good point, now return an `async` function

packages/integration-test/src/setup.js
thomas-waite: At this point, `this.contractPromises` is an array of arrays, where each inner array contains a promise, like so:

```
this.contractPromises = [ [ 'ACE', Promise { <pending> } ],
  [ 'Dividend', Promise { <pending> } ],
  [ 'ERC20Mintable', Promise { <pending> } ],
  [ 'FactoryAdjustable201907', Promise { <pending> } ],
  [ 'JoinSplit', Promise { <pending> } ],
  [ 'JoinSplitFluid', Promise { <pending> } ],
  [ 'PrivateRange', Promise { <pending> } ],
  [ 'PublicRange', Promise { <pending> } ],
  [ 'Swap', Promise { <pending> } ],
  [ 'ZkAssetAdjustable', Promise { <pending> } ] ]
```
In the `.map()`,  I'm using `Promise.all()` on each inner array (i.e. each `currentContractPromise`). My understanding is that `Promise.all()` is then a promise, which gets resolved when the promise in the inner array resolves. 

Therefore, at this stage `this.contractPromises` will be an array of promises. I need to wait for all these promises to resolve, so call `await Promise.all()` on line 153

ArnSch: No, that's not correct, and actually you point to something I missed in review:

`this.contractPromises` is an array of arrays of form `[ [ 'contractName', Promise < > ], ... ]`. Each iteration of the map will apply the `fn` provided on one of the inner arrays (`[ 'contractName', Promise ]`). You need to await that array's index 1.

Using `Promise.all` here happens to work, because `Promise.all` takes in an array of promises, and the first element of the array you give it happens to be a value (awaiting a value will just return that value), but this should be changed for clarity.

thomas-waite: Sure, that makes sense. Have changed

ArnSch (approved):

PR #384, opened 2019-11-19
Update contract interfaces

thomas-waite:
## Summary
This PR updates the previously existing contract interfaces for `ACE.sol` and the various `ZkAsset`s. It also adds interfaces for `ZkAsset`s that did not yet have them.


ArnSch (approved):
LGTM but we should automate next sprint @thomas-waite

PR #404, opened 2019-12-16
Only call `getFirstAddress()` when needed

thomas-waite:
## Summary
This PR refactors the `truffle-config.js` in the `protocol` package so that the `getFirstAddress()` function is only called when needed. This prevents unintended JSON-RPC errors.

## Description
`getFirstAddress()` pings a local node to get an Ethereum address, which is necessary for the local subproviders that enable error stack tracing, gas profiling and coverage. 

However at the moment, the `getFirstAddress()` is called regardless of the environment in which the `truffle-config` is used. This has been resulting in unexpected `JSON-RPC` errors (due to ganache not being running) when you try to call `truffle compile` for example.

## Checklist:


ArnSch (approved):
👍

PR #411, opened 2019-12-19
Feat upgradeable account registry

thomas-waite:
## Summary
This PR makes the `AccountRegistry.sol` contract, first introduced on the `feat-sdk-ui` branch upgradeable. It makes use of the unstructured proxy storage pattern developed by OpenZeppelin.

## Description
The `AccountRegistry.sol` contract is used to register register Ethereum accounts that will use the AZTEC SDK.  It needs to be upgradeable to account for the case where new methods are added to register users, but also to account for entirely new behaviour if the contract purpose evolves. 

### Architecture
The `AccountRegistry` contract has now been broken out into two contracts. There are `Behaviour` contracts, which define the methods of the behaviour and a storage contract in the form of a `Proxy` - this maintains state. The `Behaviour` contracts are the mutable, upgradeable aspect of the system.

In addition, there is now an `AccountRegistryManager` contract, which has two key methods:
- `deployProxy()`
- `upgradeAccountRegistry()` 

which are used to codify the process of deploying the proxy contract and also to upgrade the current account registry implementation. The use of this contract reduces the scope for human error in performing these actions. 

It should be noted that in the pattern implemented here, `Behaviour` contracts have to be deployed manually - there is no concept of `Factory` contracts like there is in our upgradeable note registry work. 

### Versioning system
A basic versioning system for the `Behaviour` contracts has been implemented. A basic `epoch` based system has been used, where each `Behaviour` contract is assigned an integer value. 

The system is initialised with an `epoch` of `0`, and then each `Behaviour` contract has an incremented epoch. Each behaviour contract must inherit in order from behaviour contracts with a lower epoch number, and it is only possible to upgrade to contracts with a higher epoch than the `latestEpoch` - in order to prevent any potential corruption of storage. 

Unlike upgradeable note registries, the `AccountRegistry` contracts are `assetType` and `cryptoSystem` agnostic - there is no need to account for these variables in the versioning system.

## Testing instructions
Various tests for the pattern have been added in `protocol/test/AccountRegistry`.

## Types of changes
Implement upgradeable account registries


packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * @dev Deploy the proxy contract. This stores all state for the AccountRegistry contract system
+     * @param initialBehaviourAddress - address of the behaviour contract to be linked to be initialised
+     * when the proxy is deployed 
+     * @param aceAddress - address of ACE
+     * @param trustedGSNSignerAddress - address that is being trusted to produce signatures to approve 
+     * relayed GSN calls
+     */
+    function deployProxy(
```
ArnSch: We should never be able to deploy a new Proxy contract, so think nicest way to ensure it only gets called once is to make it the constructor.

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
```diff
@@ -0,0 +1,107 @@
...
+
+        address proxyAddress = address(new AdminUpgradeabilityProxy(
+            initialBehaviourAddress,
+            adminAddress,
+            initialiseData 
+        ));
+    
+        proxy = proxyAddress;
+        behaviour = IAccountRegistryBehaviour(proxyAddress);
+        accountRegistry = initialBehaviourAddress;
```
ArnSch: I think these variables should be collapsed into 1 (just the proxy address).

You currently don't use `accountRegistry` anywhere, and in the upgrade flow, you pass in the `proxyAddress` (and so don't read from `proxy`). 

I also find it slightly odd if any callers to the account registry need to go through the manager contract (i.e. having the `behaviour` variable). Can we get away with simply storing the address of the proxy, and delegating the responsibility of handling the interface to any calling contract?

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
```diff
@@ -0,0 +1,107 @@
...
+        emit IncrementLatestEpoch(latestEpoch);
+    }
+
+    /**
+     * @dev Upgrade the account registry to a new behaviour implementation
+     * @param proxyAddress - address of the proxy contract
+     * @param newBehaviourAddress - address of the behaviour contract to be upgraded to
+    */
+    function upgradeAccountRegistry(
+        address proxyAddress,
```
ArnSch: Linked to comment above, this shouldn't be passed in, but should read from the state of the contract.

packages/protocol/contracts/AccountRegistry/TransactionRelayer.sol
```diff
@@ -0,0 +1,49 @@
...
+import "../libs/NoteUtils.sol";
+import "../interfaces/IERC20Mintable.sol";
+import "../interfaces/IZkAsset.sol";
+
+contract TransactionRelayer is IAZTEC, Initializable {
+    using NoteUtils for bytes;
+
+    ACEModule.ACE ace;
+
+    function initialize(address _ace) public initializer {
```
ArnSch: Just sense checking that it's ok for this to be callable (only once) by anyone?

packages/protocol/contracts/AccountRegistry/epochs/1/Behaviour1.sol
```diff
@@ -0,0 +1,85 @@
...
+
+    event GSNTransactionProcessed(bytes32 indexed signatureHash, bool indexed success, uint actualCharge);
+
+    /**
+    * @dev Initialize the contract and set up it's state. An initialize function rather than a constructor
+    * is used to make this compatible with the upgradeability pattern
+    * @param _aceAddress - address of the AZTEC Cryptography Engine
+    * @param _trustedGSNSignerAddress - address which will produce signature to approve relayed GSN calls
+    */
+    function initialize(address _aceAddress, address _trustedGSNSignerAddress) initializer public {
```
ArnSch: Sorry if I missed, but do you test this initialisation works as expected anywhere?

packages/protocol/contracts/AccountRegistry/epochs/1/base/BehaviourBase1.sol
```diff
@@ -0,0 +1,82 @@
...
+contract BehaviourBase1 is LibEIP712 {
+    event RegisterExtension(
+        address indexed account,
+        bytes linkedPublicKey,
+        bytes spendingPublicKey 
+    );
+
+    event LogAddress(
+        address account
+    );
```
ArnSch:
```suggestion
    );
    
```

packages/protocol/contracts/AccountRegistry/epochs/1/base/BehaviourBase1.sol
```diff
@@ -0,0 +1,82 @@
...
+
+    bytes32 private constant EIP712_DOMAIN_TYPEHASH = keccak256(abi.encodePacked(EIP712_DOMAIN));
+    bytes32 private constant SIGNATURE_TYPEHASH = keccak256(abi.encodePacked(SIGNATURE_TYPE));
+
+
+    /**
+    * @dev function to validate the user is either the sender or has submitted an EIP712 signature from the 
+    * account they are registering in the AZTEC extension.
+    **/
+    function hashAZTECAccount(AZTECAccount memory _AZTECAccount) internal view returns (bytes32){
```
ArnSch:
```suggestion
    function hashAZTECAccount(AZTECAccount memory _AZTECAccount) internal view returns (bytes32) {
```

packages/protocol/contracts/AccountRegistry/epochs/1/base/BehaviourBase1.sol
```diff
@@ -0,0 +1,82 @@
...
+
+    string private constant EIP712_DOMAIN  = "EIP712Domain(string name,string version,address verifyingContract)";
+    string private constant SIGNATURE_TYPE = "AZTECAccount(address account,bytes linkedPublicKey)";
+
+    bytes32 private constant EIP712_DOMAIN_TYPEHASH = keccak256(abi.encodePacked(EIP712_DOMAIN));
+    bytes32 private constant SIGNATURE_TYPEHASH = keccak256(abi.encodePacked(SIGNATURE_TYPE));
+
+
+    /**
+    * @dev function to validate the user is either the sender or has submitted an EIP712 signature from the
```
ArnSch: This function doesn't seem to do any validation.

packages/protocol/contracts/AccountRegistry/epochs/1/base/BehaviourBase1.sol
```diff
@@ -0,0 +1,82 @@
...
+        bytes memory _spendingPublicKey,
+        bytes memory _signature
+    ) public {
+        address signer = recoverSignature(
+            hashAZTECAccount(AZTECAccount(_account, _linkedPublicKey)),
+            _signature
+        );
+        require(_account == signer, 'signer must be the account');
+        accountMapping[_account] = _linkedPublicKey;
+        // emit event EventService
```
ArnSch: ?

packages/protocol/contracts/AccountRegistry/epochs/1/base/BehaviourBase1.sol
```diff
@@ -0,0 +1,82 @@
...
+        address _account,
+        bytes memory _linkedPublicKey,
+        bytes memory _spendingPublicKey,
+        bytes memory _signature
+    ) public {
+        address signer = recoverSignature(
+            hashAZTECAccount(AZTECAccount(_account, _linkedPublicKey)),
+            _signature
+        );
+        require(_account == signer, 'signer must be the account');
```
ArnSch: Anyone can replay this signature, and emit a new arbitrary `_spendingPublicKey`. Can this be used to grief legitimate users? Why not include the _spendingPublicKey in the hash?

Would replays/superfluous events mess with UX/UI @LeilaWang @joeandrews ?

packages/protocol/contracts/AccountRegistry/interfaces/IAccountRegistryBehaviour.sol
```diff
@@ -0,0 +1,37 @@
+pragma solidity >=0.5.0 <0.6.0;
+
+
```
ArnSch:
```suggestion
```

packages/protocol/contracts/AccountRegistry/interfaces/IAccountRegistryBehaviour.sol
```diff
@@ -0,0 +1,37 @@
...
+        address account
+    );
+
+    event LogString(
+        string message
+    );
+
+    event LogBytes(
+        bytes32 sig
+    );
```
ArnSch, line 28: We could clean up the base contract by having it extend this interface. No strong opinions though.

packages/protocol/test/AccountRegistry/AccountRegistryManager.js
```diff
@@ -0,0 +1,233 @@
...
+                await manager.upgradeAccountRegistry(existingProxy, testBehaviour.address);
+
+                const postUpgradeProxy = await manager.proxy.call();
+                expect(postUpgradeProxy).to.equal(existingProxy);
+
+                const newBehaviourAddress = await manager.getImplementation.call(existingProxy);
+                const expectedNewBehaviourAddress = testBehaviour.address;
+                expect(newBehaviourAddress).to.equal(expectedNewBehaviourAddress);
+
+                const isNewFeatureImplemented = 'newFeature()' in testBehaviour.methods;
```
ArnSch: This just tests that the contract you're using as a behaviour has the new method you put on that same contract. Would be nice to see some calls to the proxy which more explicitly test that upgrades/initialisations have happened correctly.

packages/protocol/test/AccountRegistry/AccountRegistryManager.js
```diff
@@ -0,0 +1,233 @@
...
+                const sig = r + s + v;
+
+                await behaviour.registerAZTECExtension(address, linkedPublicKey, spendingPublicKey, sig);
+
+                // perform upgrade, and confirm that registered linkedPublicKey is still present in storage
+                const testBehaviour = await TestBehaviour.new();
+                const preUpgradeProxy = await manager.proxy.call();
+                await manager.upgradeAccountRegistry(preUpgradeProxy, testBehaviour.address);
+
+                const storedLinkedPublicKey = await behaviour.accountMapping.call(address);
```
ArnSch: This is completely avoiding the Proxy contract, only testing that the deployed behaviour is setting state (in itself, not in proxy context) correctly.

packages/protocol/test/AccountRegistry/AccountRegistry.js
```diff
@@ -0,0 +1,87 @@
...
+
+contract('AccountRegistry', (accounts) => {
+    describe('Success States', () => {
+        let registryContract;
+
+        beforeEach(async () => {
+            registryContract = await AccountRegistry.new({ from: accounts[0] });
+        });
+
+        it('be able to register the extension with a valid signature', async () => {
```
ArnSch:
```suggestion
        it('should be able to register the extension with a valid signature', async () => {
```

ArnSch (changes requested):
Some updates needed around tests, and I think we can clean up the behaviour/manager a bit.

packages/protocol/test/AccountRegistry/AccountRegistry.js
thomas-waite: Done

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
thomas-waite: Good idea, have done

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
thomas-waite: Good point, done

packages/protocol/contracts/AccountRegistry/epochs/1/base/BehaviourBase1.sol
thomas-waite: Done

packages/protocol/contracts/AccountRegistry/epochs/1/base/BehaviourBase1.sol
thomas-waite: Looks like the docs are out of date with what the function does. Have updated

packages/protocol/contracts/AccountRegistry/epochs/1/base/BehaviourBase1.sol
thomas-waite: Have tidied up and removed the comment

packages/protocol/contracts/AccountRegistry/epochs/1/base/BehaviourBase1.sol
thomas-waite: This is a good point, and it shouldn't be possible to replay signatures. 

Will check with @joeandrews  but also want to include the `_spendingPublicKey` in the hash

thomas-waite: Have added signature replay protection - now checks that the signature hasn't been seen before. 

`_spendingPublicKey` can't be easily mixed into the hash as it's generated at a different/later stage to the signature

packages/protocol/contracts/AccountRegistry/interfaces/IAccountRegistryBehaviour.sol
thomas-waite: Good idea, done

packages/protocol/test/AccountRegistry/AccountRegistryManager.js
thomas-waite: Good point, have updated the tests

packages/protocol/test/AccountRegistry/AccountRegistryManager.js
thomas-waite: Yep agreed. Have now adapted that test to call a `newFeature()` function on the upgraded behaviour contract via the `proxyContract`

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
thomas-waite: Agreed, have removed the unnecessary variables

packages/protocol/contracts/AccountRegistry/TransactionRelayer.sol
thomas-waite: Have made the visibility modifiers for all `initialize()` methods, of contracts called by the Behaviour contract's `initialize()` function, `internal`. This works as the `Behaviour` contract is derived from these contracts. 

It stops any external party calling `initialize()` on `TransactionRelayer()` and supplying a false `_ace` address for example.

packages/protocol/contracts/AccountRegistry/epochs/1/Behaviour1.sol
thomas-waite: Nope, this doesn't appear to have been originally tested. 

I've added in a test that checks the initialisation is working - we check that the appropriate event is emitted.

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
ArnSch: I don't think you've fully removed them: you still declare `accountRegistry` and `behaviour`, and set `accountRegistry` in `upgradeAccountRegistry` (think that's unnecessary given you never read from it).

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
```diff
@@ -0,0 +1,104 @@
...
+ * @dev Manager contract that manages the deployment of the proxy contract and upgrading the
+ * account registry implementation.
+ *
+ * Copyright Spilsbury Holdings Ltd 2019. All rights reserved.
+ */
+contract AccountRegistryManager is Ownable, Modifiers {
+    using SafeMath for uint256;
+
+    IAccountRegistryBehaviour public behaviour;
+    address public accountRegistry;
```
ArnSch:
```suggestion
```

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
```diff
@@ -0,0 +1,104 @@
...
+ * @author AZTEC
+ * @dev Manager contract that manages the deployment of the proxy contract and upgrading the
+ * account registry implementation.
+ *
+ * Copyright Spilsbury Holdings Ltd 2019. All rights reserved.
+ */
+contract AccountRegistryManager is Ownable, Modifiers {
+    using SafeMath for uint256;
+
+    IAccountRegistryBehaviour public behaviour;
```
ArnSch:
```suggestion
```

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
```diff
@@ -0,0 +1,104 @@
...
+        
+        uint256 newBehaviourEpoch = IAccountRegistryBehaviour(newBehaviourAddress).epoch();
+        require(
+            newBehaviourEpoch > latestEpoch, 
+            'expected new registry to be of epoch greater than existing registry'
+        );
+
+        ProxyAdmin(proxyAddress).upgradeTo(newBehaviourAddress);
+        
+        accountRegistry = newBehaviourAddress;
```
ArnSch:
```suggestion
```

packages/protocol/test/AccountRegistry/AccountRegistry.js
```diff
@@ -0,0 +1,92 @@
...
+                keccak256('0x01'),
+                keccak256('0x0'),
+                sig,
+                { from: accounts[2] },
+            );
+            expect(registerExtensionReceipt.status).to.equal(true);
+        });
+    });
+
+    describe('Failure States', () => {
```
ArnSch, line 62: No replay protection test?

packages/protocol/test/AccountRegistry/AccountRegistryManager.js
```diff
@@ -0,0 +1,206 @@
...
+                const logs = await web3.eth.getPastLogs({ address: manager.address, topics: [topic] });
+
+                const deployedProxyAddress = await manager.proxyAddress.call();
+                const proxyAdmin = manager.address;
+                expect(logs.length).to.equal(1);
+                expect(parseInt(logs[0].topics[1], 16)).to.equal(parseInt(deployedProxyAddress, 16));
+                expect(parseInt(logs[0].topics[2], 16)).to.equal(parseInt(proxyAdmin, 16));
+            });
+
+            it('should set epoch number on proxy', async () => {
```
ArnSch:
```suggestion
            it('should set epoch number on manager', async () => {
```

packages/protocol/test/AccountRegistry/AccountRegistryManager.js
```diff
@@ -0,0 +1,206 @@
...
+                const { address, linkedPublicKey, spendingPublicKey, sig } = createSignature(proxyAddress);
+
+                let proxyContract = await Behaviour20200106.at(proxyAddress);
+                await proxyContract.registerAZTECExtension(address, linkedPublicKey, spendingPublicKey, sig);
+
+                // perform upgrade, and confirm that registered linkedPublicKey is still present in storage
+                const testBehaviour = await TestBehaviour.new();
+                await manager.upgradeAccountRegistry(testBehaviour.address);
+
+                proxyContract = await TestBehaviour.at(proxyAddress);
```
ArnSch: Minor, but why the reassignment?

ArnSch (changes requested):
A couple of minor things, and waiting on initialisation pattern sense check.

packages/protocol/test/AccountRegistry/AccountRegistry.js
thomas-waite: Sure, have added

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
thomas-waite: Yep, not sure why they're still there. Have now removed

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
thomas-waite: removed

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
thomas-waite: removed

packages/protocol/contracts/AccountRegistry/AccountRegistryManager.sol
thomas-waite: removed

packages/protocol/test/AccountRegistry/AccountRegistryManager.js
thomas-waite: done

packages/protocol/test/AccountRegistry/AccountRegistryManager.js
thomas-waite: There's no need for it, have removed

ArnSch (approved):
LGTM!

PR #412, opened 2019-12-19
Add missing package watch commands

thomas-waite:
## Summary
This PR adds a `watch` command in the `package.json` files of the `contract-artifacts` and `contract-addresses` packages. 

This is necessary to have the packages automatically build when `yarn watch` is run from the root during development periods.


ArnSch (approved):
LGTM

PR #428, opened 2020-01-16
Update account registry

LeilaWang:
## Summary

- integrate `TransactionRelayer` into behaviour contract
- validate `msgSender()` in `deposit()`

## Description

An extra parameter **`_AZTECaddress`** is required in function `registerAZTECExtension`. This address along with the private key of a user's aztec account could be used to send the transaction via GSN.  They are different from the (privKey, address) pair in the user's MetaMask account. This will avoid an additional MetaMask popup asking the user to sign the transaction when using GSN.

In `deposit()`, we check if `msgSender()` is equal to the given ERC20 owner. If they are the same, the sender is allowed to deposit into any account.

If `msgSender()` is different from the ERC20 owner, it has to be the owner's `_AZTECaddress`. If not, the transaction will revert. If they are the same, the sender is only allowed to deposit into the ERC20 owner's account.

`_AZTECaddress` can also be used to validate `msgSender()` for all future relayed functions.

## Testing instructions

Tests can be found in `protocol/test/AccountRegistry/AccountRegistryRelayerFunctionality`.

The tests run on contract `TestAccountMapping`, which inherits from the behaviour contract, but has an additional method to set the `_AZTECaddress` manually. This will allow us to test `deposit()` without having to send `registerAZTECExtension` first. 

## Types of changes

Validate `msgSender()` properly when sending transaction via GSN.


joeandrews (approved):
Great!

ArnSch (approved):
LGTM

